PR #14, opened 2022-09-12
IE-127 Address Sonar issues in lfp-pay-api

adr1ancho:


kdobija1 (approved):

assets/penalty_types.yml
```diff
@@ -102,7 +102,7 @@ allowed_transactions:
       true
     NN:
       true
-    NO:
+    'NO':
```
mreis-ch, line 105: Should this have quotes as none of the others do?

adr1ancho: I could quote the others too

This one is quoted and I am hoping that Sonar would recognise this literal is a token, rather than a boolean truthy value

mreis-ch: Now this has been merged to we know if by putting quotes on this particular penalty type still works ok?

PR #33, opened 2024-12-13
SAN-73 Genericise lfp specific logic in api

acallaghan-ch:
Genericise lfp specific logic
Add yaml file to load different penalty types


acallaghan-ch:
I will create a new pull request for this

PR #34, opened 2024-12-16
SAN-73: Genericise lfp specific logic

acallaghan-ch:
Genericise lfp specific logic

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,8 @@
 package config
 
 import (
+	"gopkg.in/yaml.v2"
+	"os"
 	"sync"
 	"time"
 
@@ -29,6 +31,12 @@ type Config struct {
 	PlannedMaintenanceEnd      string       `env:"PLANNED_MAINTENANCE_END_TIME"   flag:"planned-maintenance-end-time"    flagDesc:"The time of the day at which Planned E5 maintenance ends"`
 }
 
+// PenaltyDetailsMap defines the struct to hold the penalty details
+type PenaltyDetailsMap struct {
+	Description string                       `yaml:"name"`
+	Details     map[string]map[string]string `yaml:"details"`
+}
+
 // Get returns a pointer to a Config instance
 // populated with values from environment or command-line flags
 func Get() (*Config, error) {
@@ -47,4 +55,20 @@ func Get() (*Config, error) {
 	}
 
 	return cfg, nil
+}
+
+func LoadPenaltyDetails(fileName string) (*PenaltyDetailsMap, error) {
+	yamlFile, err := os.ReadFile(fileName)
+	if err != nil {
+		return nil, err
+	}
+
+	var penaltyDetailsMap PenaltyDetailsMap
+
+	err = yaml.Unmarshal(yamlFile, &penaltyDetailsMap)
+	if err != nil {
+		return nil, err
+	}
+
+	return &penaltyDetailsMap, nil
 }
--- a/config/keys.go
+++ b/config/keys.go
@@ -4,10 +4,8 @@ package config
 type ContextKey string
 
 const (
-	// DAOService is the key to store the dao service
-	DAOService = ContextKey("DAOService")
-	// CompanyNumber is the key that stores the company number
-	CompanyNumber = ContextKey("CompanyNumber")
+	// CompanyDetails is the key that stores the company number and company code
+	CompanyDetails = ContextKey("CompanyDetails")
 	// PayableResource is the key that stores the payable resource
 	PayableResource = ContextKey("PayableResource")
 )
--- a/service/penalties.go
+++ b/service/penalties.go
@@ -38,13 +38,13 @@ func (transactionType TransactionType) String() string {
 // GetPenalties is a function that:
 // 1. makes a request to e5 to get a list of transactions for the specified company
 // 2. takes the results of this request and maps them to a format that the penalty-payment-web can consume
-func GetPenalties(companyNumber string) (*models.TransactionListResponse, ResponseType, error) {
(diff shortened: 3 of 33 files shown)


config/config.go
```diff
@@ -29,6 +31,12 @@ type Config struct {
 	PlannedMaintenanceEnd      string       `env:"PLANNED_MAINTENANCE_END_TIME"   flag:"planned-maintenance-end-time"    flagDesc:"The time of the day at which Planned E5 maintenance ends"`
 }
 
+// PenaltyDetailsMap defines the struct to hold the penalty details
+type PenaltyDetailsMap struct {
+	Description string                       `yaml:"name"`
+	Details     map[string]map[string]string `yaml:"details"`
```
dgordon-ch, line 37: Adding the comment to track it, but open question on if we can find a more elegant solution to this instead of nested maps. Wonder if there's a means of doing this with objects or different file type (json?) perhaps, or even having all the fields separated and not in map. We can address it further on CS/ROE updates anyways.

config/keys.go
```diff
@@ -4,10 +4,8 @@ package config
 type ContextKey string
 
 const (
-	// DAOService is the key to store the dao service
-	DAOService = ContextKey("DAOService")
```
dgordon-ch, line 8: Reasoning for removing this? Is it all tied up in the changes to passing in middleware to create_payable_resource.go?

service/penalties.go
```diff
@@ -121,7 +125,7 @@ func generateTransactionListFromE5Response(e5Response *e5.GetTransactionsRespons
 			return nil, err
 		}
 		listItem.IsPaid = e5Transaction.IsPaid
-		listItem.Kind = "late-filing-penalty#late-filing-penalty"
+		listItem.Kind = penaltyDetailsMap.Details[penaltyType]["ResourceKind"]
```
dgordon-ch, line 128: As we're only using the ResourceKind field, would it be preferable to work this out upstream and pass in that field only, instead of always passing in the full penaltyDetailsMap?

Or do we expect other fields within this map to be used here in future iterations of this?

transformers/payable_resource_test.go
```diff
@@ -176,3 +178,20 @@ func TestUnitPayableResourceToPaymentDetails(t *testing.T) {
 		So(response.Items[0].ProductType, ShouldEqual, "late-filing-penalty")
 	})
 }
+
+func loadPenaltyDetails() *config.PenaltyDetailsMap {
```
dgordon-ch, line 182: Do we want to grab this from somewhere (the yml file) or are we okay to keep this info local to this test class? Similar question for any test class that declares these details separately.

config/config.go
acallaghan-ch: I think with JSON we would be in the same boat. The reason I went for YAMl was that there is already a yaml file used for the 'penalty_types' and wanted to keep it consistent.I had tried to use a 'Penalty' object as the value in map but ran into issues when unmarshalling.

transformers/payable_resource_test.go
acallaghan-ch: I tried loading it from the yaml but it couldn't find the file. There is probably something I am doing wrong, but I could not figure it out. I have a unit test in for the yaml unmarshalling, so I am happy to keep this as just a test object in the class, but this is more a result of my lack of Go experience.

service/penalties.go
acallaghan-ch: I honestly don't know we will need more :) I see what you mean if it is only the one value. I am happy to keep this as is for now, for ease of following through the logic i.e. just following the penatlyDetailsMap parameter through, but I think it is something to revisit.

config/config.go
jroden-ch: If the yaml in question is the one from `penalty_details.yaml` you could do the following...
```
type PenaltyDetails struct {
	Name    string    `yaml:"name"`
	Details []Details `yaml:"details"`
}

type LP struct {
	Description        string `yaml:"Description"`
	DescriptionID      string `yaml:"DescriptionId"`
	ResourceKind       string `yaml:"ResourceKind"`
	ProductType        string `yaml:"ProductType"`
	EmailReceivedAppID string `yaml:"EmailReceivedAppId"`
	EmailFilingDesc    string `yaml:"EmailFilingDesc"`
	EmailMsgType       string `yaml:"EmailMsgType"`
}

type Details struct {
	LP LP `yaml:"LP"`
}
```

This would be used like...
```
	pd := models.PenaltyDetails{}
	...
	pd.Details[0].LP.Description = "This is a description"
	pd.Details[0].LP.DescriptionID = "6d623c778ed54dcaa96664939e734618"
	pd.Details[0].LP.ResourceKind = "some resource"

        ...
	for _, Detail := range pd.Details {
		fmt.Println(Detail.LP.Description)
	}

```

This would let you use

jroden-ch: if the object for details doesn't need to be an array/slice, you can just have `PenaltyDetails` have a `Details Details` so you can just go `pd.Detail.LP.Description`, however having Details as an array lets there be multiple details in the yaml

acallaghan-ch: So we are going to have at least another two entries in the YAML for 'CS' and 'ROE'. This is why I couldn't use the 'LP' struct as it variable if that makes sense.

config/keys.go
dgordon-ch: Okay, always weary when removing something even if it's flagged as not used :)...let's see how it goes, did wonder whether there was a dependency on it somehow in external app (api-core maybe) but don't think there is after looking about.

acallaghan-ch: I had a look to see if it was used but couldn't find anything either. Also using this as an opportunity to tidy up the code.

config/config.go
acallaghan-ch: I have updated the map to use the struct as per our conversation.

dgordon-ch (approved):

PR #35, opened 2024-12-20
SAN-176: Increase unit test coverage

acallaghan-ch:
Update the unit test coverage for api.


handlers/penalties.go
```diff
@@ -32,7 +42,7 @@ func HandleGetPenalties(penaltyDetailsMap *config.PenaltyDetailsMap) http.Handle
 			return
 		}
 
-		companyCode, err := utils.GetCompanyCodeFromVars()
+		companyCode, err := getCompanyCodeFromVars()
 		if err != nil {
```
rbull-ch, line 46: Please can you see if you can cover this `companyCode` Error 400 flow, hopefully similar to something covered above for `companyNumber`?
* https://code-analysis.platform.aws.chdev.org/code?id=uk.gov.companieshouse%3Apenalty-payment-api&pullRequest=35&selected=uk.gov.companieshouse%3Apenalty-payment-api%3Ahandlers%2Fpenalties.go&line=49

acallaghan-ch: At the moment the 'companyCode' one always returns 'LP'. When we go post first release and we are catering for 'CS' too we will be able to cover it. We will need to update the call from the web component to have the penalty reference in it for this. Good spot though.

handlers/penalties_test.go
```diff
@@ -11,20 +15,110 @@ import (
...
+				req := httptest.NewRequest("GET", "/penalties", nil)
+				vars := tc.input
+				req = mux.SetURLVars(req, vars)
+				rr := httptest.NewRecorder()
+
+				handler := HandleGetPenalties(penaltyDetailsMap, allowedTransactionsMap)
+				handler.ServeHTTP(rr, req)
+
+				if rr.Code != tc.expectedStatusCode {
+					t.Errorf("handler returned wrong status code: got %v want 200", rr.Code)
```
rbull-ch, line 74: Should this be `want 400` or is it supposed to be dynamic using the expectedStatusCode above?

handlers/penalties.go
```diff
@@ -14,16 +14,26 @@ import (
 	"github.com/gorilla/mux"
 )
 
+var getCompanyNumberFromVars = func(vars map[string]string) (string, error) {
```
rbull-ch, line 17: @jroden-ch / @acallaghan-ch Is it safe to do these 3 vars at this level - wondering whether this might cause a global var race condition perhaps? As opposed to doing what was done previously in `func HandleGetPenalties`

handlers/penalties_test.go
acallaghan-ch: It is supposed to be dynamic to use the codes

service/penalties.go
```diff
@@ -35,16 +32,21 @@ func (transactionType TransactionType) String() string {
 	return transactionTypes[transactionType-1]
 }
 
+var getTransactions = func(companyNumber string, companyCode string, penaltyDetailsMap *config.PenaltyDetailsMap, client *e5.Client) (*e5.GetTransactionsResponse, error) {
```
rbull-ch, line 35: Same var scope question here as handler?
Might be ok - one for @jroden-ch to confirm

handlers/penalties.go
acallaghan-ch: To my limited knowledge yes it is ok. These are just function aliases and I don't think they are accessible outside of the  class. The reason they were pulled out is to allow for unit testing coverage as it is not possible to mock the responses without this.

utils/http_test.go
```diff
@@ -36,23 +36,31 @@ func TestUnitWriteJSON(t *testing.T) {
...
-		So(err.Error(), ShouldEqual, "company number not supplied")
+			Convey("Then err should be nil and company number should equal "+companyNumber, func() {
+				So(result, ShouldEqual, companyNumber)
+				So(err, ShouldBeNil)
+			})
+		})
+		Convey("When no company number is provided", func() {
+			vars := map[string]string{}
+			result, err := GetCompanyNumberFromVars(vars)
+			Convey("Then err should be nil and company number should equal "+companyNumber, func() {
```
rbull-ch, line 55: Can you check this convey description - does it need the `+companyNumber` bit like the previous convey?

handlers/penalties_test.go
rbull-ch: Perfect, I'll let you decide what you want to do with the `want 200` bit

acallaghan-ch: I've updated the 200 piece, let me know if I have missed what you meant as I didn't spot it the first time.

utils/http_test.go
acallaghan-ch: Good spot, copy and paste error. Updated.

rbull-ch (approved):

service/email_service.go
```diff
@@ -129,6 +151,13 @@ func prepareKafkaMessage(emailSendSchema avro.Schema, payableResource models.Pay
 		CreatedAt:    time.Now().String(),
 	}
 
+	log.Info("penaltyDetailsMap.Details[penaltyType].EmailReceivedAppId: " + penaltyDetailsMap.Details[penaltyType].EmailReceivedAppId)
+	log.Info("messageID: " + messageID)
+	log.Info("penaltyDetailsMap.Details[penaltyType].EmailMsgType: " + penaltyDetailsMap.Details[penaltyType].EmailMsgType)
+	log.Info("string(dataBytes): " + string(dataBytes))
```
dgordon-ch, line 157: Should we be logging everything out here? Is info or debug better?

dgordon-ch: Also, what sort of info are we expecting in this string?

dgordon-ch (approved):

service/email_service.go
acallaghan-ch: Debugging log message left in by mistake. Will remove.

handlers/create_payable_resource_test.go
```diff
@@ -21,24 +21,29 @@ import (
 	. "github.com/smartystreets/goconvey/convey"
 )
 
-func serveCreatePayableResourceHandler(body []byte, service dao.Service) *httptest.ResponseRecorder {
+var companyNumber = "10000024"
+
+func serveCreatePayableResourceHandler(body []byte, service dao.Service, withAuthUserDetails bool) *httptest.ResponseRecorder {
 	path := "/company/1000024/penalties/late-filing/payable"
```
jroden-ch, line 27: Not actually an issue just an observation, You could change this to be 

`template := "/company/%s/penalties/late-filing/payable"
path := fmt.sprintf(template, companyNumber)`

This would allow you to change the company number without need to change the test, This would also allow you to run table tests to make sure that it works for more than a single company

jroden-ch (approved):
Looks good to me 🚀

PR #41, opened 2025-01-29
SAN-218 Create folder structure for Issuer Gateway

aonubeze-ch:
* Created empty directories for Issuer Gateway
* Created placeholder Go files for the new directories
* Updated Docker ignore file to whitelist 'issuer_gateway' directories

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -18,3 +18,4 @@
 !go.mod
 !go.sum
 !main.go
+!issuer_gateway
--- a/issuer_gateway/api/account_penalties.go
+++ b/issuer_gateway/api/account_penalties.go
@@ -1,0 +1,1 @@
+package api
--- a/issuer_gateway/private/generate_transaction_list_from_e5_response.go
+++ b/issuer_gateway/private/generate_transaction_list_from_e5_response.go
@@ -1,0 +1,1 @@
+package private
--- a/issuer_gateway/types/transaction_type.go
+++ b/issuer_gateway/types/transaction_type.go
@@ -1,0 +1,1 @@
+package types


penalties/handlers/healthcheck_finance.go
```diff
@@ -2,14 +2,14 @@ package handlers
 
 import (
 	"fmt"
+	"github.com/companieshouse/penalty-payment-api/penalties/config"
```
rbull-ch, line 5: Why do we need to move the existing folders under another `penalties` package?

Just thinking this might make it more confusing for people used to the current go projects which appear to be more flat.  I guess I prefer having `handlers/healthcheck_finance.go`
It will take time for me to adjust to the Go code / package shuffling.

https://github.com/search?q=org%3Acompanieshouse+language%3AGo&type=code

issuer-gateway/api/AccountPenalties.go
```diff
@@ -0,0 +1 @@
+package api
```
rbull-ch, line 1: Wondering if this is the right package if it's now under IG and should the top level folder be underscore e.g. `issuer_gateway/api`?

.dockerignore
```diff
@@ -1,20 +1,20 @@
...
+!penalties/internal/middleware
+!penalties/dao/mocks
 !models
-!service
-!spec
-!transformers
-!utils
-!validators
+!penalties/service
+!penalties/api/spec
```
rbull-ch, line 14: For me having `api` here is quite confusing when we are trying to also introduce `issuer_gateway/api` as well.

Should this stuff just remain under `spec`?

I think the idea of the IG api is to hold the Public Service methods - funcs with Capitals in Go
Although I could be misunderstanding the idea entirely.

Possibly worth pulling in @jeffnayler-ch to this PR for further comments / discussion on these packages for everyones benefit

jeffnayler-ch: I'm not sure I see the need the the penalties folder, surely this is the guts of the PPS API. Wanted to pull out the Issuer Gateway to a dedicated folder structure in a way that made it feel like a separate sub-system

dgordon-ch: I don't mind the idea of it, but would agree we probably don't need the separation for penalties specifically. It's outlined in the ticket with the illustrative structure Aaron mocked together, but yeah probably better it isn't a thing for now.

main.go
```diff
@@ -4,16 +4,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/companieshouse/penalty-payment-api/penalties/config"
```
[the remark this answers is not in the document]
dgordon-ch: Yeah, noticed in Aaron's PRs as well, we need to make call on whether we separate them at bottom of imports together or keep them inline/alphabetical but grouped. I would prefer separate at bottom of imports, especially if we have a bunch of github/ch ones.

penalties/handlers/patch_payable_resource.go
```diff
@@ -3,27 +3,27 @@ package handlers
...
-	"github.com/companieshouse/penalty-payment-api/config"
 	"github.com/companieshouse/penalty-payment-api/e5"
-	"github.com/companieshouse/penalty-payment-api/service"
-	"github.com/companieshouse/penalty-payment-api/utils"
 	"gopkg.in/go-playground/validator.v9"
 )
 
 // handleEmailKafkaMessage allows us to mock the call to sendEmailKafkaMessage for unit tests
-var handleEmailKafkaMessage = service.SendEmailKafkaMessage
+var handleEmailKafkaMessage = service2.SendEmailKafkaMessage
```
dgordon-ch, line 20: If we don't use the penalties top level folder, this wouldn't need to change right and stay as `service`?

penalties/handlers/patch_payable_resource_test.go
```diff
@@ -6,7 +6,10 @@ import (
 	"encoding/json"
 	"errors"
 	"github.com/companieshouse/penalty-payment-api-core/constants"
-	"github.com/companieshouse/penalty-payment-api/mocks"
+	"github.com/companieshouse/penalty-payment-api/penalties/config"
```
dgordon-ch, line 9: As already said on the imports, I don't mind if we do keep them in alphabetic order, or separate, but we need to be consistent with it across the changes. Like I say, might be worth @acallaghan-ch confirming what he thinks as well to align.

acallaghan-ch: package handlers

import (
	"context"
	"net/http"
	"net/http/httptest"
	"testing"
	"time"

	"github.com/companieshouse/penalty-payment-api-core/models"
	"github.com/companieshouse/penalty-payment-api/config"
	"github.com/companieshouse/penalty-payment-api/utils"

)

issuer_gateway/api/AccountPenalties.go
rbull-ch, line 1: Probably one for our Dev Clinic - API Gateway Continued Discussion later.
But  if we're creating new files under `issuer_gateway/api` then I don't think we should mess too much with convention. Which looks to be all lowercase in Companies House.
e.g. I'm thinking of having 
* `issuer_gateway/api/healthcheck_finance.go` - which would be an interface
then I'd want an implementation of that - not yet decided on convention, could be to get started with
* `issuer_gateway/api/healthcheck_finance_impl.go`

I think for this PR you should use the base files you are intending to start with on your other ticket [SAN-220](https://companieshouse.atlassian.net/browse/SAN-220) to get the folder structure committed after the dev team code freeze is lifted.

* https://github.com/companieshouse/penalty-payment-api/commit/d1af66564de036d348aa35b629cbd55e81e85838#diff-3845ce715b3dc82ba8c7ba18fcc0471a2224504cee1d79dbc5a5e670f4e24622

dgordon-ch (approved):

issuer-gateway/api/accountPenalties.go
rbull-ch, line 1: Didn't we agree to try to stick to the current Go naming conventions in Companies House - so these go files would be `snake_case.go`?

.dockerignore
```diff
@@ -18,3 +18,4 @@
 !go.mod
 !go.sum
 !main.go
+!issuer-gateway
```
rbull-ch, line 21: I thought this folder would be in `snake_case` also actually?

rbull-ch: I mentioned this because that looks to be convention for the Go file names, so I thought the same for the folder we are adding.  I've had a look at these repos you mention but they don't have folders with either kebab-case or snake_case to back up either argument - see below.  If we need the folder, then I think I prefer it in underscore to try to follow the AC in [SAN-218](https://companieshouse.atlassian.net/browse/SAN-218)?

Additionally references below for

* https://github.com/companieshouse/styleguides/blob/main/go.md
* https://go.dev/doc/effective_go#package-names

> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps.

FYI @dgordon-ch  / @jeffnayler-ch

issuer-gateway/api/accountPenalties.go
aonubeze-ch: Yes we did. Done.

.dockerignore
aonubeze-ch: Done. Changed the folder name from issuer-gateway to issuer_gateway.

rbull-ch (approved):

PR #42, opened 2025-01-29
SAN-116: Allow sanctions journey

acallaghan-ch:
Allow sanctions journey


assets/penalty_details.yml
```diff
@@ -8,4 +8,12 @@ details:
...
-    EmailMsgType:       "penalty_payment_received_email"
\ No newline at end of file
+    EmailMsgType:       "penalty_payment_received_email"
+  C1:
+    Description: "Confirmation Statement"
+    DescriptionId: "confirmation-statement"
+    ResourceKind: "confirmation-statement#confirmation-statement"
+    ProductType: "confirmation-statement"
+    EmailReceivedAppId: "penalty-payment-api.penalty_payment_received_email"
+    EmailFilingDesc: "Failure to file Confirmation Statement"
```
rbull-ch, line 18: Should this be `Failure to file confirmation statement` to match the reason we see in penalty-payment-web currently?

assets/penalty_details.yml
```diff
@@ -8,4 +8,12 @@ details:
     ProductType: "late-filing-penalty"
     EmailReceivedAppId: "penalty-payment-api.penalty_payment_received_email"
     EmailFilingDesc:    "Late filing of accounts"
-    EmailMsgType:       "penalty_payment_received_email"
\ No newline at end of file
+    EmailMsgType:       "penalty_payment_received_email"
+  C1:
+    Description: "Confirmation Statement"
+    DescriptionId: "confirmation-statement"
+    ResourceKind: "confirmation-statement#confirmation-statement"
```
rbull-ch, line 15: I wonder if this could be `penalty#sanctions` for the payment details?

assets/penalty_details.yml
```diff
@@ -8,4 +8,12 @@ details:
     ProductType: "late-filing-penalty"
     EmailReceivedAppId: "penalty-payment-api.penalty_payment_received_email"
     EmailFilingDesc:    "Late filing of accounts"
-    EmailMsgType:       "penalty_payment_received_email"
\ No newline at end of file
+    EmailMsgType:       "penalty_payment_received_email"
+  C1:
+    Description: "Confirmation Statement"
```
rbull-ch, line 13: Should this be `Sanctions Penalty Payment`?
This might influence the `DescriptionId` below in kebab case...

assets/penalty_details.yml
```diff
@@ -8,4 +8,12 @@ details:
...
+    DescriptionId: "confirmation-statement"
+    ResourceKind: "confirmation-statement#confirmation-statement"
+    ProductType: "confirmation-statement"
```
rbull-ch, line 16: `penalty`?

handlers/register.go
```diff
@@ -52,8 +51,8 @@ func Register(mainRouter *mux.Router, cfg *config.Config, svc dao.Service, penal
 	mainRouter.HandleFunc("/healthcheck/finance-system", HandleHealthCheckFinanceSystem).Methods(http.MethodGet).Name("healthcheck-finance-system")
 
 	appRouter := mainRouter.PathPrefix("/company/{company_number}/penalties/late-filing").Subrouter()
-	appRouter.HandleFunc("", HandleGetPenalties(penaltyDetailsMap, allowedTransactionsMap)).Methods(http.MethodGet).Name("get-penalties")
-	appRouter.Handle("/payable", CreatePayableResourceHandler(svc, penaltyDetailsMap, allowedTransactionsMap)).Methods(http.MethodPost).Name("create-payable")
+	appRouter.HandleFunc("/{penalty_number}", HandleGetPenalties(penaltyDetailsMap, allowedTransactionsMap)).Methods(http.MethodGet).Name("get-penalties")
```
rbull-ch, line 54: If we don't intend to be doing any filtering in the api service in this PR, we are ok to remove this `penalty_number` path param and pass in a param instead for the `penalty_ref_name` e.g. String enum `["LATE_FILING", "SANCTIONS"]`
See https://github.com/companieshouse/penalty-payment-web/blob/master/src/main/java/uk/gov/companieshouse/web/pps/util/PenaltyReference.java#L4-L5

If we intend to drop `late-filing` from the endpoint path, will the endpoint become something like `GET /company/{company_number}/penalties/ref-name/{penalty_ref_name}`
Ideally this would be a query param rather than a path param e.g. `?ref-name=LATE_FILING`

Can we remember to update the build major version to be a breaking change and also the README / schema.json please?

handlers/register.go
```diff
@@ -52,8 +51,8 @@ func Register(mainRouter *mux.Router, cfg *config.Config, svc dao.Service, penal
...
+	appRouter.Handle("/{penalty_number}/payable", CreatePayableResourceHandler(svc, penaltyDetailsMap, allowedTransactionsMap)).Methods(http.MethodPost).Name("create-payable")
```
rbull-ch, line 55: As discussed earlier, our thinking is that this POST endpoint doesn't need the `{penalty_number}` path param as this looks to exist within the body (CreatePayableResource) as `transaction_id` currently, so should be reverted.

However looking at the swagger / README - think these Payment endpoints should also have the `late-filing` bit removed so the Base URL becomes: `/company/{company_number}/penalties`?

assets/penalty_details.yml
acallaghan-ch: I am think to just update this to a generic Sanctions message as discussed earlier, maybe 'Sanctions Penalty',
   `ProductType: "sanctions"
    EmailReceivedAppId: "penalty-payment-api.penalty_payment_received_email"
    EmailFilingDesc: "Sanctions Penalty'"`
, as the 'C1' company code applies to all sanctions. Then I think we need to have a mechanism for determining the messages based on the transaction subtype. I think there is a ticket for the subtype mapping so this can be tackled then. How does that sound?

assets/penalty_details.yml
acallaghan-ch: Yeah, I'll update all this to be Sanctions.

assets/penalty_details.yml
acallaghan-ch: Was thinking just 'sanctions' but no wedded top that. @dgordon-ch What is your preference?

handlers/register.go
acallaghan-ch: Yeah, going to pass the type as discussed. 
I plan to update the url to '/company/{company_number}/penalty/{penalty_ref_type}'. @dgordon-ch any preference on this?

assets/penalty_details.yml
rbull-ch: Yep, just be aware that this will come out in the email notification here - https://github.com/companieshouse/chs-notification-api/blob/main/assets/email-templates/penalty_payment_received_email.html#L79C40-L79C58
as "Reason for penalty" - so that will differ from requirements so far compared to the Penalty summary confirmation page. Probably need to run that idea past Hamayan etc

assets/penalty_details.yml
rbull-ch: Yep, I think that's what I edited from actually - tricky business

dgordon-ch: Yeah that works, penalty-sanctions, unless we find ourselves changing it later as payments api needs something else, but we can look at that before rolling out CS in general.

assets/penalty_details.yml
dgordon-ch: I'd imagine if it corresponds to the reason on the email it's going to need different text to failure to file. Let's get the BA confirm.

assets/penalty_details.yml
acallaghan-ch: Yeah, good suggestion. What about this, to keep it uniform with the 'Late Filing', just as it doesn't mention 'payment':
`  C1:
    Description: "Sanctions Penalty"
    DescriptionId: "sanctions-penalty"
    ResourceKind: "sanctions-penalty#sanctions-penalty"
    ProductType: "sanctions-penalty"
    EmailReceivedAppId: "penalty-payment-api.penalty_payment_received_email"
    EmailFilingDesc: "Sanctions Penalty Payment"
    EmailMsgType: "penalty_payment_received_email"`

handlers/register.go
dgordon-ch: Happy with either type or name, name is potentially more encompassing that it's not just C1/LP but the full name of the pen type.

assets/penalty_details.yml
rbull-ch: @acallaghan-ch / @dgordon-ch Just noticed in the swagger spec - the `cost` model shows if that helps your search with the payment details any

```
description*	string
The english description of the cost item. Derived from description_identifier and description_values.

description_identifier*	string
The enumuration identifier of the description.
```

assets/penalty_details.yml
acallaghan-ch: We might need to drive this off the transaction subtype i.e. S1 for CS, if we want to send specific messages in the email.

assets/penalty_details.yml
acallaghan-ch: Ok cool, I'll have a look into that.

rbull-ch: GitHub is being rubbish for me - but found some enums here

- https://github.com/companieshouse/api-enumerations/blob/ac67709522abb8b1293dbe00837bbb265a0a85f2/payments.yaml#L3

service/email_service_test.go
```diff
@@ -101,6 +101,12 @@ func TestUnitPrepareKafkaMessage(t *testing.T) {
 			Definition: emailSendSchema,
 		}
 
+		mockedGetCompanyCodeFromTransaction := func(transactions []models.TransactionItem) (string, error) {
+			return "LP", nil
```
dgordon-ch, line 105: Do we need the C1 journey mocked as well?

service/payment_details_test.go
```diff
@@ -49,7 +49,7 @@ func TestUnitGetPaymentDetailsFromPayableResource(t *testing.T) {
 
 		service := &PaymentDetailsService{}
 
-		paymentDetails, responseType, err := service.GetPaymentDetailsFromPayableResource(req, &payable, penaltyDetailsMap)
+		paymentDetails, responseType, err := service.GetPaymentDetailsFromPayableResource(req, &payable, penaltyDetailsMap, "LP")
```
dgordon-ch, line 52: Any chance we can pass params into the test in go and test  C1 as well?

dgordon-ch (changes requested):

handlers/register.go
dgordon-ch: We can cover it off here https://companieshouse.atlassian.net/browse/SAN-66, interesting to find it's not in the swagger spec. Would defo remove the late-filing part when we get to it.

As for penalty/penalties, I'd leave as is now, as functionally we are retrieving all penalties for a single company. But we might need to revisit as we discussed earlier re: if we add a separate post or single get penalty end point at later date (or if it comes post SOW in Java land).

handlers/penalties_test.go
```diff
@@ -1,15 +1,16 @@
...
 	"errors"
-	"github.com/companieshouse/penalty-payment-api-core/models"
-	"github.com/companieshouse/penalty-payment-api/config"
-	"github.com/companieshouse/penalty-payment-api/service"
-	"github.com/gorilla/mux"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 
+	"github.com/companieshouse/penalty-payment-api-core/models"
```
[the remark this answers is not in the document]
acallaghan-ch: Let me have a look at this and see.

service/payment_details_test.go
acallaghan-ch: I've stuck in some table tests for other things to test the LFP & Sanctions journey. I'll have a look at the tests and update to give some better coverage.

handlers/payment_details_test.go
```diff
@@ -31,6 +31,12 @@ func serveGetPaymentDetailsHandler(payableResource *models.PayableResource) *htt
 
 func TestUnitHandleGetPaymentDetails(t *testing.T) {
 
+	mockedGetCompanyCodeFromTransaction := func(transactions []models.TransactionItem) (string, error) {
+		return "LP", nil
```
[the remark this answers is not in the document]
acallaghan-ch: I can a look.

handlers/patch_payable_resource_test.go
```diff
@@ -198,7 +198,8 @@ func TestUnitPayResourceHandler(t *testing.T) {
 			mockService := mocks.NewMockService(mockCtrl)
 			mockService.EXPECT().GetPayableResource(gomock.Any(), gomock.Any()).Return(dataModel, nil)
 			mockService.EXPECT().UpdatePaymentDetails(dataModel).Times(1)
-			mockService.EXPECT().SaveE5Error("", "123", e5.CreateAction).Return(errors.New(""))
+			// commenting out for now, still looking at why this call is not made
```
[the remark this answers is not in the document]
acallaghan-ch: This is what I mentioned this morning around commenting out this mock; actively looking at it.

handlers/create_payable_resource_test.go
```diff
@@ -176,7 +181,7 @@ func TestUnitCreatePayableResourceHandler(t *testing.T) {
 		res := serveCreatePayableResourceHandler(body, mocks.NewMockService(mockCtrl), true)
 		companyNumber = "10000024"
 
-		So(res.Code, ShouldEqual, http.StatusInternalServerError)
+		So(res.Code, ShouldEqual, http.StatusBadRequest)
```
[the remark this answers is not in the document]
acallaghan-ch: Possibly, let me have a look and see.

handlers/create_payable_resource.go
```diff
@@ -42,11 +43,15 @@ func CreatePayableResourceHandler(svc dao.Service, penaltyDetailsMap *config.Pen
...
-			log.ErrorR(r, fmt.Errorf("company not in context"))
-			m := models.NewMessageResponse("company number not in request context")
+		// Determine the CompanyCode from the penaltyNumber which should be on the path
+		vars := mux.Vars(r)
+		penaltyNumber := vars["penalty_number"]
+		companyCode, err := getCompanyCode(penaltyNumber)
+
+		if err != nil {
+			log.ErrorR(r, fmt.Errorf("company code cannot be determined"))
+			m := models.NewMessageResponse("company code not not on request")
```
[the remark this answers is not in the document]
acallaghan-ch: Probably more specific as the CompanyNumber is put on the request context in the middleware. The CompanyCode is determined from the transaction list which is different.

assets/penalty_details.yml
dgordon-ch: Yeah, don't mind having that implementation as separate ticket or under the CS email one we already have in backlog (think it's still there...).

handlers/create_payable_resource.go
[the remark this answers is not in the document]
dgordon-ch: Sounds good, just clean up the 2nd not and we're golden :)

handlers/penalties_test.go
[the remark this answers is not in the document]
dgordon-ch: Align with Augustine on his SAN-218 work also.

assets/penalty_details.yml
acallaghan-ch: @rbull-ch / @dgordon-ch  Do you think we should update the 'api-enumerations' as part of this ticket by adding an entry for:
'sanctions-penalty'                     : 'Sanctions penalty'
?

handlers/create_payable_resource_test.go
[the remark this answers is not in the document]
acallaghan-ch: Actually just looking at this class and I think that it should be returning 400s up until it tries to create the resource. 500s have been used in places but for me it's actually a 400. I'll give you a shout on Slack on this.

handlers/payment_details_test.go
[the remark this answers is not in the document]
acallaghan-ch: I've put in a table test now to have both the LFP and Sanctions journeys covered.

handlers/penalties_test.go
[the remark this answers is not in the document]
acallaghan-ch: Looking at doing this structure:
`package handlers

import (
	"context"
	"net/http"
	"net/http/httptest"
	"testing"
	"time"

	"github.com/companieshouse/penalty-payment-api-core/models"
	"github.com/companieshouse/penalty-payment-api/config"
	"github.com/companieshouse/penalty-payment-api/utils"

)`

service/email_service_test.go
acallaghan-ch: I'll wrap all these tests in a table and then we'll get the two journeys

handlers/register.go
acallaghan-ch: @rbull-ch / @dgordon-ch I'm going to just update the swagger docs to reflect what the urls actually are i.e. with the late-filing bit. Then when the ticket to change these urls is being tackled we can update them all.

utils/reference.go
```diff
@@ -52,37 +55,49 @@ func GetCompanyNumberFromVars(vars map[string]string) (string, error) {
...
-	if len(penaltyNumber) == 0 {
-		return "", fmt.Errorf("penalty number not supplied")
+// GetCompanyCode gets the company code from the penalty reference type
+func GetCompanyCode(penaltyReferenceType string) (string, error) {
+	if len(penaltyReferenceType) == 0 {
+		return "", fmt.Errorf("penalty reference type not supplied")
 	}
 
-	penaltyPrefix := penaltyNumber[0]
+	log.Info(" *****************: " + penaltyReferenceType)
```
dgordon-ch, line 64: Do we want this?

dgordon-ch (approved):

handlers/penalties.go
```diff
@@ -4,21 +4,18 @@ import (
...
 	"github.com/companieshouse/penalty-payment-api/utils"
-	"github.com/gorilla/mux"
 )
 
-var getCompanyNumberFromVars = func(vars map[string]string) (string, error) {
-	return utils.GetCompanyNumberFromVars(vars)
-}
-var getCompanyCodeFromVars = func() (string, error) {
-	return utils.GetCompanyCodeFromVars()
+var getCompanyCode = func(penaltyReference string) (string, error) {
```
rbull-ch, line 17: Looks like this needs to be renamed to be `penaltyReferenceType`?

handlers/penalties.go
```diff
@@ -29,30 +26,20 @@ func HandleGetPenalties(penaltyDetailsMap *config.PenaltyDetailsMap, allowedTran
 	return func(w http.ResponseWriter, req *http.Request) {
 		log.InfoR(req, "start GET penalties request from e5")
 
-		// Check for a company number in request
-		vars := mux.Vars(req)
-		// only company number in the route variables
-
-		companyNumber, err := getCompanyNumberFromVars(vars)
+		companyNumber := req.Context().Value(config.CompanyNumber).(string)
```
rbull-ch, line 29: Are we forgetting to handle this missing `company_number` path param - Bad Request error now?

handlers/penalties.go
acallaghan-ch: Will update.

handlers/penalties.go
acallaghan-ch: As this is taken from the context, which is populated in a middleware unction which is ran when the request comes in, the error handling for the 'Bad Request' lives in there instead of having a duplicate here. 
https://github.com/companieshouse/penalty-payment-api/blob/7c0d5085cda135815de0ccea76fbf237fdb61ff2/handlers/register.go#L61
https://github.com/companieshouse/penalty-payment-api/blob/7c0d5085cda135815de0ccea76fbf237fdb61ff2/middleware/company.go#L21

rbull-ch: Apologies, think I noticed this after I commented

spec/schema.json
```diff
@@ -6,7 +6,7 @@
     "description": "An API that creates, patches, and retrieves Penalty Payment Service (PPS) sessions, as well as interacting with the E5 finance system"
   },
   "host": "api.companieshouse.gov.uk",
-  "basePath": "/company/{company_number}/penalties",
+  "basePath": "/company/{company_number}/penalties/late-filing",
```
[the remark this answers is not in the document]
acallaghan-ch: I'll remove the 'basepath' in that case and just define the full url in the  path definitions.

spec/schema.json
```diff
@@ -69,12 +69,15 @@
             "in": "path",
             "type": "string",
             "required": true
+          },
+          {
+            "name": "penalty_reference_type",
+            "in": "path",
+            "type": "string",
```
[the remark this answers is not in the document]
acallaghan-ch: That's not available in our version of Swagger unfortunately, but I have included an inline enum.
https://swagger.io/docs/specification/v2_0/enums/

spec/schema.json
```diff
@@ -200,7 +215,18 @@
...
             }
+          },
+          "400": {
+            "description": "Bad request - Invalid input"
+          },
+          "404": {
+            "description": "payable resource does not exist or has insufficient data"
+          },
+          "500": {
+            "description": "payable resource does not exist or has insufficient dat"
```
rbull-ch, line 226: Minor typo here.
If you're going to correct this, can you update the `/info/version` in the swagger at the same time?

Also the `version` file for the build, so I'd expect the major version to be bumped in both if the GET penalties endpoint doesn't handle the old late-filing request without the new path param.
Otherwise, I guess this version would just be a minor version bump.

acallaghan-ch: I will fix the typo and update the major version. I think it is better we don't cater for the old-request without a parameter and get other teams to update when we are going to release.

rbull-ch: Ok, just a bit confused why we changed the `api-sdk-java` regex yesterday to accept the old GET path if we don't intend to support it in our new version of the API.  So this will affect other services when we move into cidev later this sprint. If @dgordon-ch is ok with that, then I guess he can liase with the other team leads to get that integration work done.  But they might gripe a bit when we come to further refactoring it another time to drop the `late-filing` from the path.

acallaghan-ch: Until we release this it should be ok, and I think we will have further refactoring of that regex as you say to drop the 'late-filing'. The update from yesterday allows us the scope to support the old way if it is decided that we must, i.e. default to LATE_FILNG, but for me that would be the wrong option and they should be forced to update. However, if we are told we must support the old way it will just mean an api change.

rbull-ch: Still don't see [version](https://github.com/companieshouse/penalty-payment-api/blob/master/version) in the list of changes

acallaghan-ch: @dgordon-ch / @rbull-ch Should I default to 'LATE_FILIN' for now if a request comes in on the old URL to save making waves for now? Then when we get agreement on the services using the API regarding updating their URL we can remove this default behaviour.

utils/reference_test.go
```diff
@@ -42,10 +44,170 @@ func TestUnitGenerateEtag(t *testing.T) {
...
+						So(err, ShouldBeNil)
+					}
+					So(companyCode, ShouldEqual, tc.expectedCode)
+				})
+			})
+		}
+	})
+}
+
+func TestUnitGetCompanyCodeFromTranaction(t *testing.T) {
```
rbull-ch, line 140: Should this be `Transaction`?

utils/reference_test.go
```diff
@@ -42,10 +44,170 @@ func TestUnitGenerateEtag(t *testing.T) {
...
+						Amount:        5,
+						Type:          "penalty",
+						TransactionID: "Q1000007",
+					},
+				},
+				expectedCode:  "",
+				expectedError: true,
+			},
+			{
+				name: "Error no tranaction ID",
```
rbull-ch, line 184: Should this be `Transaction`?

utils/reference_test.go
acallaghan-ch: Will update

utils/reference_test.go
acallaghan-ch: Will update

handlers/create_payable_resource.go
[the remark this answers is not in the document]
rbull-ch: Still got the `not not` here @acallaghan-ch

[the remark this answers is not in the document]
rbull-ch: Also wondering whether the message should be more similar to the log you have above - some mention of "cannot be determined" rather than what we have here?

spec/schema.json
acallaghan-ch: Going to default it for now, let me know if any other preference.

handlers/create_payable_resource.go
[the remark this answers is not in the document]
acallaghan-ch: I think you are right, it should be "cannot be determined"

version
```diff
@@ -1 +1 @@
-0.4
+2.0.0
```
rbull-ch, line 1: It might be less confusing in the git tag history if this was 1 major version jump?
`1.0.0`?

rbull-ch (approved):

PR #45, opened 2025-02-05
Feature/austin/sanctions branch

acallaghan-ch:
Merge feature branch into master


spec/schema.json
```diff
@@ -1,12 +1,11 @@
 {
   "swagger": "2.0",
   "info": {
-    "version": "1.0.0",
+    "version": "0.5.0",
```
rbull-ch, line 4: I think we should just set this schema version to be `1.1.0` since we had 1.0 before.
It doesn't need to be kept in line with the build tag as it will get out of step that way.

dgordon-ch (approved):

spec/schema.json
acallaghan-ch: Ok, I will update this. I was getting confused between the build tag and the api version, I thought that they should match up but I get what you are saying now.

rbull-ch (approved):

PR #46, opened 2025-02-06
SAN-219 Update HandleHealthCheckFinanceSystem to use new IssuerGatewayHealthcheckFinanceSystem - CheckScheduledMaintenance

rbull-ch:
[SAN-219](https://companieshouse.atlassian.net/browse/SAN-219) Update HandleHealthCheckFinanceSystem to use new IssuerGatewayHealthcheckFinanceSystem - CheckScheduledMaintenance
* Add unit tests
* Add Named Function Returns to improve readability
* Update returnWeeklyMaintenanceTime to strip seconds, milliseconds and monotonic clock
* Update swagger GET /healthcheck/finance-system - 503 Service Unavailable response example


jeffnayler-ch (approved):

issuer_gateway/api/healthcheck_finance.go
```diff
@@ -0,0 +1,9 @@
+package api
+
+import (
+	"time"
+)
+
+type HealthcheckFinanceSystem interface {
+	CheckScheduledMaintenance() (systemAvailableTime time.Time, systemUnavailable bool, parseError bool)
+}
```
acallaghan-ch, line 9: We might not need this as an interface:
https://stackoverflow.com/questions/57063010/golang-multi-method-interface-and-struct-naming
It saying that if there is only one method in the interface then it probably isn't needed. They do seem to go for the naming convention *Impl in Go either.

rbull-ch: Yep agree - done in commit [ef528dc](https://github.com/companieshouse/penalty-payment-api/commit/ef528dc493ae34ea5ed580ec0e4885a6d1c88e25)

issuer_gateway/api/healthcheck_finance_test.go
```diff
@@ -0,0 +1,437 @@
...
+	"github.com/companieshouse/penalty-payment-api/config"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestUnitCheckScheduledMaintenance(t *testing.T) {
+	cfg, _ := config.Get()
+	currentTime := time.Now()
+
+	Convey("No maintenance config", t, func() {
+		// Given
```
dgordon-ch, line 17: Do we need this if it's blank (although I do like the structure/idea behind it). Nothing major and can leave as is.

dgordon-ch (approved):

acallaghan-ch (approved):

PR #50, opened 2025-02-14
SAN-294 Set confirmation email paid date to current date

aonubeze-ch:


dgordon-ch (approved):
Agree with this approach due to the transaction date not necessarily having up to date information (or expected date) at time we retrieve it for email service to use. Can revisit when we implement the cached/db for transactions as that would in theory be working source of truth.

service/email_service.go
```diff
@@ -120,18 +120,13 @@ func prepareKafkaMessage(emailSendSchema avro.Schema, payableResource models.Pay
 		err = fmt.Errorf("error parsing made up date: [%v]", err)
 		return nil, err
 	}
-	transactionDate, err := time.Parse("2006-01-02", payedTransaction.TransactionDate)
```
rbull-ch, line 123: Please can you correct the comment on line 117, since we've removed `transactionDate`?
Maybe just use the 2 field names on the `DataField`?

rbull-ch (approved):

service/email_service.go
aonubeze-ch: Great catch, thanks.

PR #51, opened 2025-02-17
SAN-230 Remove EmailFilingDesc config and use new transactionListItem.Reason field in API (via IG api.AccountPenalties)

rbull-ch:
[SAN-230](https://companieshouse.atlassian.net/browse/SAN-230) Remove `EmailFilingDesc` config and use new `transactionListItem.Reason` field in API (via IG `api.AccountPenalties`)

[SAN-230]: https://companieshouse.atlassian.net/browse/SAN-230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/email_service.go
+++ b/service/email_service.go
@@ -129,7 +129,7 @@ func prepareKafkaMessage(emailSendSchema avro.Schema, payableResource models.Pay
 		TransactionDate:   time.Now().Format("2 January 2006"),
 		Amount:            fmt.Sprintf("%g", payedTransaction.OriginalAmount),
 		CompanyName:       companyName,
-		FilingDescription: penaltyDetailsMap.Details[companyCode].EmailFilingDesc,
+		FilingDescription: payedTransaction.Reason,
 		To:                payableResource.CreatedBy.Email,
 		Subject:           fmt.Sprintf("Confirmation of your Companies House penalty payment"),
 		CHSURL:            cfg.CHSURL,
--- a/assets/penalty_details.yml
+++ b/assets/penalty_details.yml
@@ -7,7 +7,6 @@ details:
     ResourceKind: "late-filing-penalty#late-filing-penalty"
     ProductType: "late-filing-penalty"
     EmailReceivedAppId: "penalty-payment-api.penalty_payment_received_email"
-    EmailFilingDesc: "Late filing of accounts"
     EmailMsgType: "penalty_payment_received_email"
   C1:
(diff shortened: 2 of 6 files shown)


acallaghan-ch (approved):

service/email_service.go
```diff
@@ -129,7 +129,7 @@ func prepareKafkaMessage(emailSendSchema avro.Schema, payableResource models.Pay
 		TransactionDate:   time.Now().Format("2 January 2006"),
 		Amount:            fmt.Sprintf("%g", payedTransaction.OriginalAmount),
 		CompanyName:       companyName,
-		FilingDescription: penaltyDetailsMap.Details[companyCode].EmailFilingDesc,
+		FilingDescription: payedTransaction.Reason,
```
dgordon-ch, line 132: Can we count on this reason value to correspond to the desired "Failure to file a confirmation statement"? I thought we'd still need to pass that value in here as we can't rely on data, is it worth posing this to Hamayun/Naz?

dgordon-ch: I'm more getting at, the response data from e5 might not always hold that reason, so if we don't map it anywhere on our side wouldn't you just get whatever value they're holding in that field (which from experience could be something that doesn't correspond to "Failure to file..."?

That payedTransaction object, is getting the penalties (saying it gets one but it's making the same get all call to e5...) and whatever you have in that field, it'll put it in that filing description.

Agree on the new `api.AccountPenalties` call and yes for now I'd keep it there in service package.

acallaghan-ch: Yeah that was hardcoded for now to get the first pass at the CS journey. With the sub-type mapping stuff coming in I would expect that value to be pulled from the yaml file. I would probably update the Late Filing one too just to make it consistent, probably driving them off reading the companyCode and Subtype: ex. penaltyDetailsMap.Details[companyCode].Subtype[subType].filingDescription

dgordon-ch: Spoke to Richard, think we're okay as this reason value is being updated in the `GenerateTransactionListFromE5Response.getTransactionType` when it's pulling back the penalty info from e5. I was under false assumption this reason wasn't being set there based on the if/else around LP/C1.

Happy to use this reason and remove filing description if it's not longer required.

dgordon-ch (approved):

PR #52, opened 2025-02-18
SAN-272 Token for payment in CS

klai-ch:
SAN-272 CS - Making a payment to the correct GovPay account by passing in expected token

### DIFF
--- a/assets/penalty_details.yml
+++ b/assets/penalty_details.yml
@@ -8,6 +8,7 @@ details:
     ProductType: "late-filing-penalty"
     EmailReceivedAppId: "penalty-payment-api.penalty_payment_received_email"
     EmailMsgType: "penalty_payment_received_email"
+    ClassOfPayment: "penalty"
   C1:
     Description: "Sanctions Penalty Payment"
     DescriptionId: "penalty-sanctions"
@@ -15,3 +16,4 @@ details:
     ProductType: "penalty-sanctions"
     EmailReceivedAppId: "penalty-payment-api.penalty_payment_received_email"
     EmailMsgType: "penalty_payment_received_email"
+    ClassOfPayment: "data-maintenance"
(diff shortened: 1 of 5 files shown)


assets/penalty_details.yml
```diff
@@ -8,10 +8,12 @@ details:
     ProductType: "late-filing-penalty"
     EmailReceivedAppId: "penalty-payment-api.penalty_payment_received_email"
     EmailMsgType: "penalty_payment_received_email"
+    ClassOfPayment: "penalty"
```
rbull-ch, line 11: Please could you keep these with the other 4 `cost` details like we talked about earlier?
Looking at the model, I think we should have `ClassOfPayment` after `DescriptionId`.

```
cost{
description*	[...]
description_identifier*	[...]
description_values*	{...}
amount*	[...]
available_payment_methods*	[...]
class_of_payment*	[...]
kind*	[...]
resource_kind*	[...]
product_type*	[...]
}
```

transformers/payable_resource.go
```diff
@@ -112,7 +112,7 @@ func PayableResourceToPaymentDetails(payable *models.PayableResource,
 		cost := models.Cost{
 			Amount:                  fmt.Sprintf("%g", tx.Amount),
 			AvailablePaymentMethods: []string{"credit-card"},
-			ClassOfPayment:          []string{"penalty"},
+			ClassOfPayment:          []string{penaltyDetailsMap.Details[companyCode].ClassOfPayment},
```
rbull-ch, line 115: It's a pity we didn't just pass in the `PenaltyDetails` as a function parameter rather than having to do this Map stuff each time.

rbull-ch (approved):

assets/penalty_details.yml
klai-ch: Update as suggested from review

acallaghan-ch (approved):

transformers/payable_resource.go
acallaghan-ch: Ask David to create a refactor ticket and it can be updated.

klai-ch: Refactor is created for this issue
https://companieshouse.atlassian.net/browse/SAN-301

klai-ch:
This PR can then help QAs potentially when trying to test the account routing with chBearer (using data-maintenance). And there will be another ticket for the new mapping (penalty-sanctions, penalty-lfp) and related dependencies. 
https://companieshouse.atlassian.net/browse/SAN-303

PR #53, opened 2025-02-21
SAN-303 Penalty Payment Mappings

klai-ch:
SAN-303 Add new payment mappings for ClassOfPayment in payments-api and dependencies


service/payment_details_test.go
```diff
@@ -188,3 +188,178 @@ func TestUnitGetPaymentDetailsFromPayableResource(t *testing.T) {
 
 	})
 }
+
+func TestUnitGetPaymentDetailsFromPayableResourceSanctions(t *testing.T) {
```
acallaghan-ch, line 192: I think that we could use a table test for this instead of duplicating the tests for sanctions. Have a look at : handlers/create_payable_resource_test.go
It would also mean that when we add other types in future the tests would be easy to extend and we would have less code duplication.

acallaghan-ch (commented):
Possible test changes

dgordon-ch (approved):

PR #54, opened 2025-02-21
SAN-224 Use Issuer Gateway PayablePenalty API in Send Email service

aonubeze-ch:
* SendEmailKafkaMessage calls Issuer Gateway PayablePenalty in place of GetTransactionForPenalty.
* Unit test refactored to reflect the update.
* Refactored PayablePenalty/MatchPenalty to process a single transaction

### DIFF
--- a/issuer_gateway/api/payable_penalty.go
+++ b/issuer_gateway/api/payable_penalty.go
@@ -11,8 +11,8 @@ import (
 var getAccountPenalties = AccountPenalties
 var getMatchingPenalty = private.MatchPenalty
 
-func PayablePenalty(companyNumber string, companyCode string, txs []models.TransactionItem,
-	penaltyDetailsMap *config.PenaltyDetailsMap, allowedTransactionsMap *models.AllowedTransactionMap) ([]models.TransactionItem, error) {
+func PayablePenalty(companyNumber string, companyCode string, transaction models.TransactionItem,
+	penaltyDetailsMap *config.PenaltyDetailsMap, allowedTransactionsMap *models.AllowedTransactionMap) (*models.TransactionItem, error) {
 
 	response, _, err := getAccountPenalties(companyNumber, companyCode, penaltyDetailsMap, allowedTransactionsMap)
 	if err != nil {
@@ -27,10 +27,10 @@ func PayablePenalty(companyNumber string, companyCode string, txs []models.Trans
 			"company_number": companyNumber,
 			"penalty_count":  unpaidPenaltyCount,
 		})
-		return []models.TransactionItem{}, private.ErrMultiplePenalties
+		return nil, private.ErrMultiplePenalties
 	}
 
-	return getMatchingPenalty(response.Items, txs, companyNumber)
+	return getMatchingPenalty(response.Items, transaction, companyNumber)
 }
 
 func getUnpaidPenaltyCount(transactionListItems []models.TransactionListItem) int {
--- a/service/email_service.go
+++ b/service/email_service.go
@@ -72,18 +72,12 @@ func SendEmailKafkaMessage(payableResource models.PayableResource, req *http.Req
 }
 
 var getCompanyCodeFromTransaction = utils.GetCompanyCodeFromTransaction
-
 var getCompanyName = GetCompanyName
-
 var getPayablePenalty = api.PayablePenalty
 
 // prepareKafkaMessage generates the kafka message that is to be sent
-func prepareKafkaMessage(emailSendSchema avro.Schema,
-	payableResource models.PayableResource,
-	req *http.Request,
-	penaltyDetailsMap *config.PenaltyDetailsMap,
-	allowedTransactionsMap *models.AllowedTransactionMap) (*producer.Message, error) {
-
+func prepareKafkaMessage(emailSendSchema avro.Schema, payableResource models.PayableResource, req *http.Request,
+	penaltyDetailsMap *config.PenaltyDetailsMap, allowedTransactionsMap *models.AllowedTransactionMap) (*producer.Message, error) {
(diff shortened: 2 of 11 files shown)


service/email_service.go
```diff
@@ -106,30 +101,34 @@ func prepareKafkaMessage(emailSendSchema avro.Schema, payableResource models.Pay
...
+		payableResource.CompanyNumber,
+		companyCode,
+		payableResource.Transactions,
+		penaltyDetailsMap,
+		allowedTransactionsMap)
 	if err != nil {
 		err = fmt.Errorf("error getting transaction for penalty: [%v]", err)
 		return nil, err
 	}
+	paidTransaction := payablePenalty[0]
```
acallaghan-ch, line 115: Could the program get here and it have no payable penalties in the list? Ids there a check further back that means there is always 1 penalty in the list, or do we need to check the list is not empty before accessing it here?

service/email_service.go
```diff
@@ -41,7 +42,6 @@ func SendEmailKafkaMessage(payableResource models.PayableResource, req *http.Req
 		return err
 	}
 
-	// Get a producer
 	kafkaProducer, err := getProducer(cfg)
```
acallaghan-ch, line 45: Why have you removed all the comments?

service/email_service.go
```diff
@@ -72,29 +71,25 @@ func SendEmailKafkaMessage(payableResource models.PayableResource, req *http.Req
...
-var getTransactionForPenalty = func(companyNumber, companyCode, penaltyReference string, penaltyDetailsMap *config.PenaltyDetailsMap,
-	allowedTransactionsMap *models.AllowedTransactionMap) (*models.TransactionListItem, error) {
-	return GetTransactionForPenalty(companyNumber, companyCode, penaltyReference, penaltyDetailsMap, allowedTransactionsMap)
-}
+var getPayablePenalty = api.PayablePenalty
 
 // prepareKafkaMessage generates the kafka message that is to be sent
-func prepareKafkaMessage(emailSendSchema avro.Schema, payableResource models.PayableResource, req *http.Request,
-	penaltyDetailsMap *config.PenaltyDetailsMap, allowedTransactionsMap *models.AllowedTransactionMap) (*producer.Message, error) {
+func prepareKafkaMessage(emailSendSchema avro.Schema,
```
acallaghan-ch, line 81: Why have all the function arguments been moved to separate lines?

service/email_service.go
```diff
@@ -141,7 +140,10 @@ func prepareKafkaMessage(emailSendSchema avro.Schema, payableResource models.Pay
 		return nil, err
 	}
 
-	messageID := "<" + payableResource.Reference + "." + strconv.Itoa(util.Random(0, 100000)) + "@companieshouse.gov.uk>"
+	messageID := "<" +
```
acallaghan-ch, line 143: Why has the layout been changed here?

acallaghan-ch (changes requested):

service/email_service.go
aonubeze-ch: Do we really need them? The comments seem redundant as the functions are intuitive enough, like

// Get a producer
getProducer()

service/email_service.go
aonubeze-ch: PayablePenalty returns a list with at least one item, otherwise, it returns error. This is implicit though. I am working on refactoring this. Would be something to discuss further.

service/email_service.go
aonubeze-ch: The lines are too long. It is true we do not currently enforce max characters per line but it will be nice to not have to scroll horizontally to be able to see the end of a line.

service/email_service.go
aonubeze-ch: The lines are too long. It is true we do not currently enforce max characters per line but it will be nice to not have to scroll horizontally to be able to see the end of a line.

service/email_service.go
acallaghan-ch: That's grand, just wanted to check.

service/email_service.go
acallaghan-ch: I would just create the new line once the limit has been hit, but not have a new line for each.

service/email_service.go
acallaghan-ch: I would do the same here, just create a new line once the limit has been hit but not a new line for each.

service/email_service.go
acallaghan-ch: Fair enough.

service/email_service.go
aonubeze-ch: Done.

service/email_service.go
aonubeze-ch: Done.

dgordon-ch (approved):

issuer_gateway/api/payable_penalty.go
```diff
@@ -27,10 +27,10 @@ func PayablePenalty(companyNumber string, companyCode string, txs []models.Trans
 			"company_number": companyNumber,
 			"penalty_count":  unpaidPenaltyCount,
 		})
-		return []models.TransactionItem{}, private.ErrMultiplePenalties
+		return nil, private.ErrMultiplePenalties
```
acallaghan-ch, line 30: I would have left it returning the empty List here instead of passing the 'nil' reference. The it would mean that you could just check the return to see if it has items instead of having to check to see if it is 'nil' and also non-empty. Better not to pass nils or nulls around.

aonubeze-ch: The return type changed from list to a struct, so that's why I am returning nil.

acallaghan-ch (approved):

PR #56, opened 2025-02-27
SAN-237 Update get-penalties to add a payable status to indicate the current state of the payable transaction from E5

rbull-ch:
[SAN-237](https://companieshouse.atlassian.net/browse/SAN-237) Update get-penalties to add a payable status to indicate the current state of the payable transaction from E5
    * Add DunningStatus to e5 Transaction

[SAN-237]: https://companieshouse.atlassian.net/browse/SAN-237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


issuer_gateway/private/generate_transaction_list_from_e5_response.go
```diff
@@ -84,3 +85,21 @@ func getReason(transaction *e5.Transaction) string {
...
 }
+
+const (
+	OpenPayableStatus   = "OPEN"
+	ClosedPayableStatus = "CLOSED"
+)
+
+func getPayableStatus(transaction *e5.Transaction) string {
+	if transaction.IsPaid || transaction.OutstandingAmount <= 0 || (transaction.AccountStatus == "DCA" && transaction.DunningStatus == "DCA") {
+		return ClosedPayableStatus
```
acallaghan-ch, line 96: Do we need to check the 'AccountStatus' here, I thought it was just the dunningStatus?

rbull-ch: Yes, it need to be both.  See

- https://companieshouse.atlassian.net/wiki/spaces/NS/pages/5130256450/When+a+penalty+can+t+be+paid
- https://companieshouse.atlassian.net/wiki/spaces/NS/pages/5129732167/When+a+penalty+can+be+paid

acallaghan-ch: I can that in the table, but I thought from conversations that it was really only the 'dunningStatus' that mattered. From reading the logic, I could have my Account not in 'DCA' but my 'dunningStatus' in DCA and it would be classed as 'OPEN'. I don't think this is right.

dgordon-ch: Yep as Aaron says, think only scenario where it sort of matters is CS PEN3, but we'd need to check that specifically (if type CS and dunning PEN3) and it's not tied to accountStatus. Can run it by Hamayun/Naz once more to confirm.

rbull-ch: Fixed in commit [0500354](https://github.com/companieshouse/penalty-payment-api/pull/56/commits/0500354c66d6031ff34cf3a8723c1c01fbb42046).

Thanks, I think I did have it using only DCA Dunning Status initially, but then changed it when seen the rule having both fields as DCA.
I was initially looking to swap out a few conditions from the -web to the -api which was previously only concerned with DCA Account Status.
I know we'll be deep diving further into the payable rules for LateFilingPenalty and Sanctions in the other tickets.

aonubeze-ch (approved):

dgordon-ch (approved):

PR #58, opened 2025-03-03
SAN-164 Sanctions Payable status is only Open for Account status [DCA/CHS/HLD] and Dunning status [PEN1, PEN2]

rbull-ch:
[SAN-164](https://companieshouse.atlassian.net/browse/SAN-164) Sanctions Payable status is only Open for Account status [DCA/CHS/HLD] and Dunning status [PEN1, PEN2]
* Update transactionListItem.IsDCA to use dunning status is dca instead of account status

[SAN-164]: https://companieshouse.atlassian.net/browse/SAN-164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


issuer_gateway/private/generate_transaction_list_from_e5_response.go
```diff
@@ -105,7 +108,9 @@ func getPayableStatus(transaction *e5.Transaction) string {
 
 	if transaction.CompanyCode == utils.LateFilingPenalty {
 		return OpenPayableStatus
-	} else if transaction.CompanyCode == utils.Sanctions {
+	} else if transaction.CompanyCode == utils.Sanctions &&
+		(transaction.DunningStatus == Pen1DunningStatus || transaction.DunningStatus == Pen2DunningStatus) &&
+		(transaction.AccountStatus == DcaAccountStatus || transaction.AccountStatus == ChsAccountStatus || transaction.AccountStatus == HldAccountStatus) {
 		return OpenPayableStatus
```
acallaghan-ch, line 114: This is looks grand based on the table in the ticket. This is just from memory but I thought that finance said that as long ad the dunning status was in the 'PEN1' or 'PEN2' status then the penalty could be paid, the account status shouldn't matter. Might be worth checking with Nazeeta and Hamayun.

rbull-ch: @acallaghan-ch  Waiting on a reply from Hamayan...

Hamayun - Fri 28 Feb

> Helloo, as long as the dunning status is in DCA, then we shouldn't allow the user to continue to pay
> the CS and LFP tickets in sprint 9 will outline when we can pay
> anything else we basically can't pay

Richard - Mon 3 Mar

> Hi Hamayan - what do you make of this comment from Aaron please https://github.com/companieshouse/penalty-payment-api/pull/58#discussion_r1977743713
> I have added Account Status to our sanctions condition for OPEN payable status.
> 
> That means these would be CLOSED payable status
> * Sanctions with outstanding amount, not paid, account status is wdr, dunning status is pen1
> * Sanctions with outstanding amount, not paid, account status is wdr, dunning status is pen2
> * Sanctions with outstanding amount, not paid, account status is wdr, dunning status is pen3
> 
> Not sure if these are even pen3 dunning status can happen, but if they can then they will also be CLOSED payable status
> * Sanctions with outstanding amount, not paid, account status is dca, dunning status is pen3
> * Sanctions with outstanding amount, not paid, account status is chs, dunning status is pen3
> * Sanctions with outstanding amount, not paid, account status is hld, dunning status is pen3

issuer_gateway/private/generate_transaction_list_from_e5_response.go
```diff
@@ -93,9 +93,12 @@ const (
 	ChsAccountStatus = "CHS"
 	DcaAccountStatus = "DCA"
 	HldAccountStatus = "HLD"
+	WdrAccountStatus = "WDR"
 
 	DcaDunningStatus  = "DCA"
```
aonubeze-ch, line 98: What do you think about following this Go recommendation here https://go.dev/wiki/CodeReviewComments, specifically **Initialisms** section:
* ChsAccountStatus => CHSAccountStatus
* DcaAccountStatus => DCAAccountStatus
* HldAccountStatus => HLDAccountStatus
* WdrAccountStatus => WDRAccountStatus
* DcaDunningStatus => DCADunningStatus

issuer_gateway/private/generate_transaction_list_from_e5_response.go
```diff
@@ -93,9 +93,12 @@ const (
...
 	Pen1DunningStatus = "PEN1"
+	Pen2DunningStatus = "PEN2"
+	Pen3DunningStatus = "PEN3"
```
aonubeze-ch, line 101: These ones too.

issuer_gateway/private/generate_transaction_list_from_e5_response.go
rbull-ch: You're too kind :) Done in commit [87c08dd](https://github.com/companieshouse/penalty-payment-api/pull/58/commits/87c08dd452c2b2df5c526bbb876a87be82438764)

issuer_gateway/private/generate_transaction_list_from_e5_response.go
rbull-ch: Done in commit [87c08dd](https://github.com/companieshouse/penalty-payment-api/pull/58/commits/87c08dd452c2b2df5c526bbb876a87be82438764)

issuer_gateway/private/generate_transaction_list_from_e5_response.go
aonubeze-ch: Based on the ACs, I think the implementation is correct. The ACs explicitly define open status, but closed status can be inferred from that definition. My understanding is that any combination other than those defined are considered closed payable status. If something needs to change, then it should be first, the ACs and then the implementation based on the ACs.

aonubeze-ch (approved):

issuer_gateway/private/generate_transaction_list_from_e5_response.go
dgordon-ch: Yeah I think Aaron is right, in sense the account status could basically have anything in it and they'd still want to accept payment if dunning is PEN1/2, but that's not what's on ticket, so we should follow the ACs there unless otherwise stated.

If we find later that WDR, or any old gibberish can be put in account status, then we can always remove this validation.

dgordon-ch (approved):

issuer_gateway/private/generate_transaction_list_from_e5_response.go
rbull-ch: @acallaghan-ch  / @dgordon-ch  / @aonubeze-ch Thanks all.  Hamayan has got back to me and said

> yes that's bang on

My thought is that if we want to relax the rules around payable later, it is much easier to specify what can't be paid first, rather than what we have in the tickets currently.

PR #64, opened 2025-03-06
SAN-256: Sonarqube upgrade

dnair-ch:


go.mod
```diff
@@ -1,72 +1,79 @@
 module github.com/companieshouse/penalty-payment-api
 
-go 1.19
+go 1.23.0
```
mike-kelly-ch, line 3: Don't need the minor version. `go 1.23` will suffice.

PR #80, opened 2025-03-21
Refactor(SAN-324): Merge for relocation of code to create two distinct implementations

jroden-ch:
As part of SAN-324, it was agreed that a structure update would occur to create distinctions between Penalty payments and the Issuer gateway

The idea is to create a common components, an issuer gateway and penalty payments package(s) to separate out code but have as much reusability as possible

*Originally a draft PR #77 which closed automatically when I rebased*


acallaghan-ch (changes requested):

handlers/patch_payable_resource.go
```diff
@@ -3,6 +3,9 @@ package handlers
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/companieshouse/penalty-payment-api/common/e5"
+	"github.com/companieshouse/penalty-payment-api/common/utils"
+	service2 "github.com/companieshouse/penalty-payment-api/penalty_payments/service"
```
rbull-ch, line 8: Can we remove this `service2`?  I suppose it was done to help with the initial refactoring work like @acallaghan-ch  has already commented on the `e6` stuff.  Can we tidy these loose ends up please

jroden-ch: image
This was because of the package name being similar to the other package in common. Reorganised and removed though:

image

rbull-ch (approved):

PR #84, opened 2025-03-27
SAN-266 Update Payable resource and Financial penalty models

aonubeze-ch:
Update Payable resource and Financial penalty models
* Reference updated to PayableReference for PayableResource
* TransactionID/TransactionReference updated to PenaltyRef for Transaction and FinancialPenalty

Linked PRs:
https://github.com/companieshouse/api-sdk-java/pull/397
https://github.com/companieshouse/api-sdk-java/pull/399
https://github.com/companieshouse/chs-notification-api/pull/520
https://github.com/companieshouse/penalty-payment-web/pull/267


common/dao/mongo.go
```diff
@@ -65,19 +65,19 @@ type MongoService struct {
...
 		return err
 	}
 
 	filter := bson.M{"_id": dao.ID}
 	update := bson.D{
 		{
 			Key: "$set", Value: bson.D{
-				{Key: "e5_command_error", Value: string(action)},
-			},
+			{Key: "e5_command_error", Value: string(action)},
```
rbull-ch, line 79: Can you select these lines and add a tab please to correct the indentation for readability please?

common/dao/mongo.go
```diff
@@ -143,11 +143,11 @@ func (m *MongoService) UpdatePaymentDetails(dao *models.PayableResourceDao) erro
 	update := bson.D{
 		{
 			Key: "$set", Value: bson.D{
-				{Key: "data.payment.status", Value: dao.Data.Payment.Status},
-				{Key: "data.payment.reference", Value: dao.Data.Payment.Reference},
-				{Key: "data.payment.paid_at", Value: dao.Data.Payment.PaidAt},
-				{Key: "data.payment.amount", Value: dao.Data.Payment.Amount},
-			},
+			{Key: "data.payment.status", Value: dao.Data.Payment.Status},
```
rbull-ch, line 146: Can you select these lines and add a tab please to correct the indentation for readability please?

common/dao/mongo.go
aonubeze-ch: Thought I fixed that already. Thanks. I will fix it up.

aonubeze-ch: Thought I fixed that already. Thanks. I will fix it up.

go.sum
```diff
@@ -27,8 +27,10 @@ github.com/companieshouse/gofigure v0.1.0/go.mod h1:W7NZ4kQdDXqvqK1f7EKTc+pnnqpF
 github.com/companieshouse/gofigure v0.1.4 h1:BN7Dqn0xvNHwC+KHOUHigwP7ajkfHORos3hKRtEACUU=
 github.com/companieshouse/gofigure v0.1.4/go.mod h1:lBAsI13q21rMce+pO7x4xUYAXzO0MchuDsZUbCs0Yng=
 github.com/companieshouse/htmlform v0.1.0/go.mod h1:fWScPVYjGyeBhYwppGaGRDtCQ5hMd3t3Q+0I6HaUZT4=
-github.com/companieshouse/penalty-payment-api-core v1.6.2 h1:IpE0kwX2yUuE5T857vyqP67tWuBzqeK6RSCUAPqaogA=
-github.com/companieshouse/penalty-payment-api-core v1.6.2/go.mod h1:9cqGa5ZlpwYqFh5NlqIJJK39G4p2wBdoWJkuVW8Ssxc=
+github.com/companieshouse/penalty-payment-api-core v1.8.0 h1:2LbeHM9WPjCkon0mkKaBH9lgbWMunsP3xoYvsvbG908=
```
rbull-ch, line 30: Can you run Go > Mod Tidy on the `go.mod` file please to remove these `penalty-payment-api-core v1.8.0` versions?

handlers/patch_payable_resource.go
```diff
@@ -60,15 +60,15 @@ func PayResourceHandler(payableResourceService *services.PayableResourceService,
 		err = v.Struct(request)
 
 		if err != nil {
-			log.ErrorR(r, err, log.Data{"payable_reference": resource.Reference, "payable_id": request.Reference})
+			log.ErrorR(r, err, log.Data{"payable_ref": resource.PayableRef, "payable_id": request.Reference})
```
rbull-ch, line 63: Need to be careful around the logging here, it looks to me like `payable_id` should be `payment_reference`?
So thinking we need to correct here and what about the schema.json swagger PatchResourceRequest shows reference currently.  Is that wrong?
Should we update the model here as well https://github.com/companieshouse/penalty-payment-api-core/blob/v1.9.0/models/request_entities.go#L33

handlers/patch_payable_resource.go
```diff
@@ -42,16 +42,16 @@ func PayResourceHandler(payableResourceService *services.PayableResourceService,
 		resource := i.(*models.PayableResource)
 
 		log.Info("processing penalty payment", log.Data{
-			"payable_reference": resource.Reference,
-			"customer_code":     resource.CustomerCode,
+			"payable_ref":   resource.PayableRef,
+			"customer_code": resource.CustomerCode,
 		})
 
 		// 2. validate the request and check the reference number against the payment api to validate that it has
```
rbull-ch, line 49: Should `2. validate the request and check the reference number` be `2. validate the request and check the payment reference`?

handlers/patch_payable_resource_test.go
```diff
@@ -119,7 +119,7 @@ func TestUnitPayResourceHandler(t *testing.T) {
 			So(body.Message, ShouldEqual, "no payable request present in request context")
 		})
 
-		Convey("reference is required in request body", func() {
+		Convey("payable reference is required in request body", func() {
```
rbull-ch, line 122: This looks wrong if it should be `payment reference`?

common/utils/reference_test.go
```diff
@@ -149,9 +149,9 @@ func TestUnitGetCompanyCodeFromTransaction(t *testing.T) {
 				name: "Late Filing",
 				input: []models.TransactionItem{
 					{
-						Amount:        5,
-						Type:          "penalty",
-						TransactionID: "A1000007",
+						Amount:     5,
+						Type:       "penalty",
+						PenaltyRef: "A1000007",
```
rbull-ch, line 154: Does the convey line 141 need to say `Get Company Code from penalty ref` instead?
Since we have replaced TransactionID with PenaltyRef in TransactionItem.

common/utils/reference_test.go
```diff
@@ -9,12 +9,12 @@ import (
 )
 
 func TestUnitGenerateReferenceNumber(t *testing.T) {
-	Convey("Reference Number is correct length", t, func() {
+	Convey("TransactionReference Number is correct length", t, func() {
```
rbull-ch, line 12: I think the 2 lines in the convey saying `TransactionReference Number` should be reverted

handlers/patch_payable_resource.go
rbull-ch: I think that's something different unfortunately, daos are for mongo db persistence which we haven't updated in the payment block.  My comment relates to the confusion around the request body naming and logging.

aonubeze-ch: I don't mind changing to `payment_reference` if that is what we want. I think it makes sense to do that.

issuer_gateway/api/account_update.go
```diff
@@ -32,14 +32,14 @@ func UpdateIssuerAccountWithPenaltyPaid(payableResourceService *services.Payable
...
-			Value:     t.Amount,
+			TransactionReference: t.PenaltyRef,
+			Value:                t.Amount,
 		})
 	}
 
 	// this will be used for the PUON value in E5. it is referred to as paymentId in their spec. X is prefixed to it
 	// so that it doesn't clash with other PUON's from different sources when finance produce their reports - namely
-	// ones that begin with 'LP' which signify penalties that have been paid outside of the digital service.
+	// ones that begin with 'LP' which signify penalties that have been paid outside the digital service.
```
rbull-ch, line 42: Think it reads better in the previous version?

issuer_gateway/api/account_update_test.go
```diff
@@ -78,10 +78,10 @@ func TestUnitUpdateIssuerAccountWithPenaltyPaid(t *testing.T) {
 			c := &e5.Client{}
 			p := validators.PaymentInformation{Amount: "150", PaymentID: "123"}
 			r := models.PayableResource{
-				Reference:    "123",
+				PayableRef:   "123",
 				CustomerCode: "10000024",
 				Transactions: []models.TransactionItem{
-					{TransactionID: "123", Amount: 150},
+					{PenaltyRef: "123", Amount: 150},
```
rbull-ch, line 84: Please can you update the swagger model documentation too with `penalty_ref` - https://github.com/companieshouse/penalty-payment-api/blob/SAN-266-model-refactoring/spec/schema.json#L421
i.e. FinancialPenaltySession > transactions

penalty_payments/service/company_service.go
```diff
@@ -8,17 +8,17 @@ import (
 )
 
 // GetCompanyName will attempt to get the company name from the CompanyProfileAPI.
-func GetCompanyName(companyNumber string, req *http.Request) (string, error) {
+func GetCompanyName(customerCode string, req *http.Request) (string, error) {
```
rbull-ch, line 11: I don't think it makes sense to rename `companyNumber` to `customerCode` here.
I think this file should be reverted.
Do you agree @dgordon-ch ?

dgordon-ch: Could go either way but yeah would agree, it's not going to tie into customerCode usages outside companyNumber (ROE, etc) as we'd not be hitting the company profile api to retrieve that string. As it's self-contained here, happy for it to remain as is as long as upstream still refers to customerCode where applicable.

issuer_gateway/api/account_update.go
aonubeze-ch: Both actually means the same thing but 'outside' is more commonly used than 'outside of'. I will revert anyways since you asked.

penalty_payments/interceptors/payable_resource_authentication.go
```diff
@@ -92,9 +92,9 @@ func preCheckRequest(w http.ResponseWriter, r *http.Request) (string, string, st
 		w.WriteHeader(http.StatusBadRequest)
 		return "", "", "", true
 	}
-	payableID := vars["payable_id"]
-	if payableID == "" {
-		log.InfoR(r, "PayableAuthenticationInterceptor error: no payable_id")
+	payableRef := vars["payable_ref"]
```
rbull-ch, line 95: Please can you replace the `payable_id` with `payable_ref` in the comment on line 87 and also the 3 places in the `README.md`?  That will get rid of all occurrences of the old path param, thanks.

spec/schema.json
```diff
@@ -361,7 +361,7 @@
           "items": {
             "type": "object",
             "properties": {
-              "transaction_id": {
+              "penalty_ref": {
```
rbull-ch, line 364: A general comment for visibility, but probably nothing to action here as the api changes around `_id` to `_ref` appears to be much improved to avoid confusion.  However the `penalty_ref` under transactions may need to change in the future, if we are concerned with paying other transaction types as there is a separate type field documented below on line 371.

I think it's always expected to be `"type": "penalty"` currently, so this might be a bit redundant.

rbull-ch (approved):
Approved - with 1 observation https://github.com/companieshouse/penalty-payment-api/pull/84/files#r2022558168

dgordon-ch (approved):

PR #86, opened 2025-04-02
SAN-273: Creation of DAO implementation for account penalties cache

acallaghan-ch:
Creation of DAO implementation for account penalties cache operations


common/dao/mongo.go
```diff
@@ -64,6 +64,149 @@ type MongoService struct {
 	CollectionName string
 }
 
+// MongoService is an implementation of the Service interface using MongoDB as the backend driver.
+type MongoAccountService struct {
+	db             MongoDatabaseInterface
+	CollectionName string
+}
+
```
aonubeze-ch, line 72: Is this really necessary? Can't we use the MongoService struct for both payable resource and account penalties collections? The CollectionName is a parameter so we can set the corresponding collection name when we use the struct.

common/dao/service.go
```diff
@@ -26,6 +26,29 @@ func NewDAOService(cfg *config.Config) Service {
...
+
+// NewAccountPenaltiesDaoService will create a new instance of the AccountPenaltiesCacheService interface.
+// All details about its implementation and the  database driver will be hidden from outside of this package
+func NewAccountPenaltiesDaoService(cfg *config.Config) AccountPenaltiesCacheService {
+	database := getMongoDatabase(cfg.MongoDBURL, cfg.Database)
+	return &MongoAccountService{
+		db:             database,
+		CollectionName: cfg.AccountPenaltiesCollection,
 	}
 }
```
aonubeze-ch, line 54: As you have AccountPenaltiesCacheService and NewAccountPenaltiesDaoService, I think it would be nice to rename the existing Service and NewDaoService to name specific to PayableResources. The current names are too generic and made sense when we had only one collection. What do you think?

common/dao/mongo.go
```diff
@@ -64,6 +64,149 @@ type MongoService struct {
...
+	return &resource, nil
+}
+
+// UpdateAccountPenaltyAsPaid will update the penalty status of an item in account_penalties database collection
+func (m MongoAccountService) UpdateAccountPenaltyAsPaid(customerCode string, companyCode string, penaltyReference string) error {
+	log.Info("updating penalty as paid in account_penalties collection", log.Data{
+		"customer_code":              customerCode,
+		"company_code":               companyCode,
+		"data.transaction_reference": penaltyReference,
```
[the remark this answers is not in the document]
acallaghan-ch: I can update.

common/dao/service.go
acallaghan-ch: I can update.

rbull-ch (approved):
Approved - I guess you could bump the version to `0.15`

PR #90, opened 2025-04-11
Feature/san 326 mongo cache ac

acallaghan-ch:


handlers/register.go
```diff
@@ -76,7 +75,7 @@ func Register(mainRouter *mux.Router, cfg *config.Config, daoService dao.Payable
 	// other routes
 	payResourceRouter := appRouter.PathPrefix("/penalties/payable/{payable_ref}/payment").Methods(http.MethodPatch).Subrouter()
 	payResourceRouter.Use(payableAuthInterceptor.PayableAuthenticationIntercept, authentication.ElevatedPrivilegesInterceptor)
-	payResourceRouter.Handle("", PayResourceHandler(payableResourceService, e5Client, penaltyDetailsMap, allowedTransactionsMap)).Name("mark-as-paid")
+	payResourceRouter.Handle("", PayResourceHandler(payableResourceService, e5Client, penaltyDetailsMap, allowedTransactionsMap, apDaoService)).Name("mark-as-paid")
```
rbull-ch, line 78: Why are we using the DAOs directly above and here?  Should we introduce a Service like `payableResourceService` instead to expose the operations instead of using `apDaoService` as a param here?  We are mixing service styles / convention here.

acallaghan-ch: apDaoService is a service itself so I don't see the need to introduce another service here. I had initially looked at doing this but thought it unnecessary after some googling on the approach. It also allows us to do everything through dependency injection instead of having to create a var in that is publicly accessible (line 19 for the PayableResourceService).

acallaghan-ch: There's a lot more happening in the PayableResource service, so I don't think we need the extra layer.

rbull-ch: Yep, I can see an example previously on `CreatePayableResourceHandler` that had used the other DaoService.
I just have a hard time following the Go code all the way through, I would expect that the service layer we have in the `issuer_gateway/api` could just make use of the DAOs directly as a package global variable where it is needed in the underlying services.

handlers/create_payable_resource_test.go
```diff
@@ -10,29 +10,28 @@ import (
...
-	"github.com/companieshouse/penalty-payment-api-core/models"
-	"github.com/companieshouse/penalty-payment-api/config"
-	"github.com/companieshouse/penalty-payment-api/mocks"
 	. "github.com/smartystreets/goconvey/convey"
 )
 
 var customerCode = "10000024"
 
-func serveCreatePayableResourceHandler(body []byte, service dao.PayableResourceDaoService, withAuthUserDetails bool) *httptest.ResponseRecorder {
+func serveCreatePayableResourceHandler(body []byte, payableResourceService dao.PayableResourceDaoService, accPenService dao.AccountPenaltiesDaoService,
```
rbull-ch, line 27: The param name seems to jump around different names between files, should we not ensure that they are all consistent between the .go and _test.go files?
Should they all be `apDaoSvc dao.AccountPenaltiesDaoService` since it is more widely used in this PR?

Then similar convention needed for the other dao service naming.

handlers/register.go
acallaghan-ch: I'm trying to avoid using the package global variable if possible as Go is trying to move away from their use. I know I introduced a load of them to mock out tests though :)

aonubeze-ch (approved):

handlers/create_payable_resource_test.go
rbull-ch: Much better thanks, I can see you've updated them now.  Only place that is different is handlers/register.go * `apDaoService dao.AccountPenaltiesDaoService`

rbull-ch (approved):
Approved - https://github.com/companieshouse/penalty-payment-api/pull/90/files/4c13525d67ade1c39243bf25615956621b8a9948#r2042443655

PR #91, opened 2025-04-14
SAN-276 Implement logic for handling stale cache record

aonubeze-ch:
Implement logic for handling stale Account Penalties cache record

Linked PRs:
https://github.com/companieshouse/docker-chs-development/pull/1697
https://github.com/companieshouse/chs-configs/pull/2710


version
```diff
@@ -1 +1 @@
-0.18
```
acallaghan-ch, line 1: Can you put the version as 0.19 - I reverted the version bump I did as part of my ticket hence why it is showing as '0.18' here.

common/dao/mongo.go
```diff
@@ -214,28 +214,50 @@ func (m *MongoAccountPenaltiesService) UpdateAccountPenaltyAsPaid(customerCode s
...
+			"created_at": dao.CreatedAt,
+			"closed_at":  dao.ClosedAt,
+			"data":       dao.AccountPenalties,
+		},
+	}
 
-	_, err := collection.DeleteOne(context.Background(), filter)
+	collection := m.db.Collection(m.CollectionName)
 
+	result, err := collection.UpdateOne(context.Background(), filter, update)
```
dgordon-ch, line 239: What happens if the filter doesn't return a document, will it return error here and we log out the customer/company codes? Is it worth being more explicit around what is failing?

common/dao/mongo.go
```diff
@@ -214,28 +214,50 @@ func (m *MongoAccountPenaltiesService) UpdateAccountPenaltyAsPaid(customerCode s
...
-	log.Info("successfully deleted document in account_penalties collection", log.Data{
-		"customer_code": customerCode,
-		"company_code":  companyCode,
-	})
+	if result.ModifiedCount == 1 {
+		log.Info("updated a document in account_penalties collection", log.Data{
+			"customer_code": dao.CustomerCode,
+			"company_code":  dao.CompanyCode,
+		})
+	} else {
```
dgordon-ch, line 253: Is there any possibility the modified count is > 1 or would we only expect 0/1 (as UpdateOne operation is being used)?

README.md
```diff
@@ -22,6 +22,7 @@ In order to run this API locally you will need to install the following:
 | `PPS_MONGODB_DATABASE`                     |   `-`   | The database name to connect to e.g. `late_filing_penalties`          |
 | `PPS_MONGODB_PAYABLE_RESOURCES_COLLECTION` |   `-`   | The collection name e.g. `payable_resources`                          |
 | `PPS_MONGODB_ACCOUNT_PENALTIES_COLLECTION` |   `-`   | The collection name e.g. `account_penalties`                          |
+| `PPS_ACCOUNT_PENALTIES_TTL_HOURS`          |   `-`   | Account penalties cache time to live  e.g. `24`                       |
```
dgordon-ch, line 25: As mentioned, created additional ticket to cover thinking about implementation on the allocation routine timings instead of a flat 24 hours period of time. For this ticket specifically though, 24 hours is a good initial step that can be refined -- see https://companieshouse.atlassian.net/browse/SAN-409.

common/dao/service.go
```diff
@@ -39,8 +39,8 @@ type AccountPenaltiesDaoService interface {
 	GetAccountPenalties(customerCode string, companyCode string) (*models.AccountPenaltiesDao, error)
 	// UpdateAccountPenaltyAsPaid will update a transactions as paid for a given customerCode, companyCode and penaltyRef
 	UpdateAccountPenaltyAsPaid(customerCode string, companyCode string, penaltyRef string) error
-	// DeleteAccountPenalties will delete the entry for a given customerCode and companyCode
```
dgordon-ch, line 42: We might still need a delete account penalties operation down the line, especially when we get to ruggedisation and cleanup of actual stale data (>7 days old or whatever we decide is stale). More note than anything...

issuer_gateway/api/account_penalties.go
```diff
@@ -21,25 +21,29 @@ var generateTransactionList = private.GenerateTransactionListFromAccountPenaltie
...
 	accountPenalties, err := apDaoSvc.GetAccountPenalties(customerCode, companyCode)
 
-	if accountPenalties == nil {
-		cfg, err := getConfig()
-		if err != nil {
-			return nil, services.Error, nil
-		}
+	cfg, err := getConfig()
+	if err != nil {
+		return nil, services.Error, nil
```
dgordon-ch, line 32: Is it worth logging this out, also we're not passing the err back and instead passing nil, would it be more prudent to pass the err back as well?

issuer_gateway/api/account_penalties.go
```diff
@@ -57,15 +61,26 @@ func AccountPenalties(customerCode string, companyCode string, penaltyDetailsMap
...
-	return &convertedResponse
+
+	return &accountPenalties
+}
+
+func updateAccountPenaltiesEntry(customerCode string, companyCode string, e5Response *e5.GetTransactionsResponse, apDaoSvc dao.AccountPenaltiesDaoService) *models.AccountPenaltiesDao {
+	accountPenalties := convertE5Response(customerCode, companyCode, e5Response)
+	err := apDaoSvc.UpdateAccountPenalties(&accountPenalties)
+	if err != nil {
+		log.Error(fmt.Errorf("error updating account penalties: [%v]", err),
```
dgordon-ch, line 79: In theory, we log out error, silently consume it and then continue to return the penalties. So, if we do fail to update, we're not bombing out, which makes sense, as user doesn't need to understand that failure and if we have the accountPenalties we can continue upstream, next time they're accessed (< 24 hours) could be where the problem is as it won't have been updated in cache (as paid) would that then cause an issue around duplicate payment?

Fine if so, to be expected, based on discussions we've had, and without adding further robustness to both sides re: failover. But one to be aware of, probably best to put new ticket in around this for future work.

dgordon-ch (changes requested):

issuer_gateway/api/account_penalties.go
aonubeze-ch: The next time, since we did not update the cache, it would be > 24 hours, so we try to update the cache again. You are right, the idea is to continue and return the penalties from E5 even if the cache update did not happen. We would have the cron job to clean up the cache eventually.

issuer_gateway/api/account_penalties.go
aonubeze-ch: Will do.

common/dao/mongo.go
acallaghan-ch: As it's UpdateOne the modifiedCount will only ever be 1 as the operation finds the first record that matches and updates. There should only be 1 entry in the DB for it.

README.md
aonubeze-ch: Yeah, this is the initial implementation. I am thinking that it would nice to have this parameter as an env var so we can easily change the value without code change.

common/dao/mongo.go
aonubeze-ch: UpdateOne modifies only one document. https://www.mongodb.com/docs/manual/reference/method/db.collection.updateOne/?msockid=0ef8c837ba3464131ca8dd2ebb136530#mongodb-method-db.collection.updateOne

common/dao/mongo.go
acallaghan-ch: There is logging on line 254 to cover a document not being found (or not being updated).

aonubeze-ch: The following code takes care of that bit: 
`if result.ModifiedCount == 1 {
		log.Info("updated a document in account_penalties collection", log.Data{
			"customer_code": dao.CustomerCode,
			"company_code":  dao.CompanyCode,
		})
	} else {
		err = errors.New("failed to update document in account_penalties collection")
		log.Error(err, log.Data{
			"customer_code": dao.CustomerCode,
			"company_code":  dao.CompanyCode,
		})
		return err
	}`
Remember, we are expecting the call to modify a single document. If it fails to modify the document, we return an error.

acallaghan-ch (changes requested):
Version change

common/dao/service.go
aonubeze-ch: Will do.

acallaghan-ch: I would leave it out for now until we actually need it. It will be handy enough to put back in but if we end up just using Mongo triggers or the like it will be code that is never called.

aonubeze-ch: Ok

issuer_gateway/api/account_penalties.go
```diff
@@ -105,3 +121,9 @@ func convertE5Response(customerCode, companyCode string, response *e5.GetTransac
 		AccountPenalties: data,
 	}
 }
+
+func isStale(accountPenaltiesDao *models.AccountPenaltiesDao, cfg *config.Config) bool {
+	cachRecordExpirationTime := accountPenaltiesDao.CreatedAt.Add(time.Hour * cfg.AccountPenaltiesTTLHours)
```
acallaghan-ch, line 126: small typo - cache
Just thinking that if the env variable is not set, should it default to 24 hours @dgordon-ch ?

aonubeze-ch: Fixed typo. Defaults TTL to 24 hours if not set in config.

issuer_gateway/api/account_penalties.go
aonubeze-ch: Done. Logging the error returned from `getConfig`

dgordon-ch (approved):

PR #95, opened 2025-04-15
SAN-278 Update payment status for cached account penalty

klai-ch:
SAN-278 Update payment status for cached account penalty
https://companieshouse.atlassian.net/browse/SAN-278


handlers/patch_payable_resource.go
```diff
@@ -94,6 +95,30 @@ func PayResourceHandler(payableResourceService *services.PayableResourceService,
 		w.WriteHeader(http.StatusNoContent) // This will not be set if status has already been set
 	})
 }
+
+func updateAccountPenaltyAsPaid(resource *models.PayableResource, svc dao.AccountPenaltiesDaoService) {
+
+	companyCode, err := getCompanyCodeFromTransaction(resource.Transactions)
+	if err != nil {
+		log.Error(fmt.Errorf("error updating penalty as paid because company code cannot be determined: [%v]", err),
```
dgordon-ch, line 103: I'd change "determined" to "resolved" or "reconciled", or maybe "validated".

@acallaghan-ch, what have you used in other areas around failed validation type logging?

handlers/patch_payable_resource.go
```diff
@@ -94,6 +95,30 @@ func PayResourceHandler(payableResourceService *services.PayableResourceService,
...
+			log.Data{"customer_code": resource.CustomerCode, "payable_ref": resource.PayableRef})
+		return
+	}
+	penalty := resource.Transactions[0]
```
dgordon-ch, line 107: Can the resource have multiple transactions at this point or are we always expecting 1? If we can have > 1, how do we know the first is the correct one?

handlers/patch_payable_resource.go
```diff
@@ -94,6 +95,30 @@ func PayResourceHandler(payableResourceService *services.PayableResourceService,
...
+
+	err = svc.UpdateAccountPenaltyAsPaid(resource.CustomerCode, companyCode, penalty.PenaltyRef)
+	if err != nil {
+		log.Error(fmt.Errorf("error updating penalty as paid: [%v]", err),
+			log.Data{"customer_code": resource.CustomerCode, "company_code": companyCode,
+				"penalty_ref": penalty.PenaltyRef, "payable_ref": resource.PayableRef})
+		return
+	}
+
+	log.Info("penalty is updated as paid",
```
dgordon-ch, line 117: Should this log message specify the account_penalties collection has been updated? I notice in the updateAsPaidInDatabase method, we have logging to say the payment resource has been updated in db, is it worth refactoring that method to be more explicit that it's updating payment_resource and this one we're adding is updating the account_penalties collections? 

Happy for @acallaghan-ch to chip in here also.

dgordon-ch (changes requested):

handlers/patch_payable_resource.go
acallaghan-ch: We use this logic in other places as we can only have one transaction. If we introduce the ability top pay multiple penalties that may change but for now know that there can be only one.

handlers/patch_payable_resource.go
acallaghan-ch: Yeah, I think it referencing the fact that it is updating the payable resource here would be helpful.

klai-ch: Updated log message from penalty is updated to account penalties collection is updated

handlers/patch_payable_resource.go
klai-ch: Updated from determined to resolved

acallaghan-ch (approved):

dgordon-ch (approved):

PR #100, opened 2025-04-24
SAN-410 Update get-penalties endpoint (AccountPenalties) to ensure that a penalty with costs associated by made_up_date is not payable

rbull-ch:
[SAN-410](https://companieshouse.atlassian.net/browse/SAN-410) Update get-penalties endpoint (AccountPenalties) to ensure that a penalty with costs associated by `made_up_date` is not payable
* `"payable_status": "CLOSED"`


dgordon-ch (approved):

acallaghan-ch (approved):

aonubeze-ch (commented):
I thought the idea here is to also include information about associated costs in the response payload for calls to GET penalties just as we now do for pending payment allocation. This way, we do not have to determine the associated costs in the frontend. Currently, with this implementation, we only know that the penalty is closed but we do not know why

PR #101, opened 2025-04-25
SAN-409 Update logic to determine stale cache record

aonubeze-ch:
Before this update, Account Penalties cache record is considered stale if it is 24 hours or more older (CreatedAt time is 24 hours less than the current time). 
This update enhances this logic for calculating stale cache record by also incorporating into the logic estimates for the E5 allocation routine run. It also includes the functionality to ensure that E5 has latest payment information for a given account before updating the cache with E5 data.

Linked PRs:
https://github.com/companieshouse/chs-configs/pull/2713
https://github.com/companieshouse/docker-chs-development/pull/1713


issuer_gateway/api/account_penalties_test.go
```diff
@@ -52,6 +51,23 @@ var page = e5.Page{
 var e5TransactionsResponse = e5.GetTransactionsResponse{
 	Page: page,
 	Transactions: []e5.Transaction{
+		{
+			CompanyCode:          utils.LateFilingPenalty,
+			LedgerCode:           "E1",
```
rbull-ch, line 56: Please can you update this to be more like the other late-filing penalty transactions?
We seem to have a confusing mix of e5 transaction fields for a sanctions penalty here
```
			LedgerCode:           "E1",
			TransactionReference: "P1234567",
			TransactionSubType:   "S1",
			TypeDescription:      "CS01",
```

issuer_gateway/api/account_penalties_test.go
```diff
@@ -172,7 +191,7 @@ var staleAccountPenalties = models.AccountPenaltiesDao{
 			MadeUpDate:           "2025-02-12",
 			Amount:               250,
 			OutstandingAmount:    250,
```
rbull-ch, line 193: I think the `OutstandingAmount` should be 0 if it is marked as `IsPaid` since we only allow payment in full

issuer_gateway/api/account_penalties.go
```diff
@@ -123,11 +123,58 @@ func convertE5Response(customerCode, companyCode string, response *e5.GetTransac
...
+		// Penalty is not yet marked as paid in cache so logic to determine if cache record is stale
+		// is based on a time to live of 24 hours or more.
+		ttlHours := cfg.AccountPenaltiesTTLHours
+		if ttlHours == 0 {
+			ttlHours = 24 // time to live defaults to 24 hours if not set in config
+		}
+		cacheRecordExpirationTime := accountPenaltiesDao.CreatedAt.Add(time.Hour * ttlHours)
+		now := time.Now()
+
+		log.Info("checking if cached account penalties have become stale", log.Data{
```
rbull-ch, line 136: Not sure about your use of `log.Data` here - if you need this in a debug map then convention appears to be lower snake_case fields.  See [cidev-application-logs](https://logging-common-kibana.aws.chdev.org/s/cidev/app/discover#/doc/cidev-application-logs-*/cidev-application-logs-2025.04.28-003235?id=Lr6Ye5YBCKxslaPe9hAk)

issuer_gateway/api/account_penalties.go
```diff
@@ -123,11 +123,58 @@ func convertE5Response(customerCode, companyCode string, response *e5.GetTransac
...
+		return cacheRecordExpirationTime.Equal(time.Now()) || cacheRecordExpirationTime.Before(now)
+	} else {
+		// Penalty is marked as paid in cache, so logic to determine if cache record is stale is based on
+		// E5 allocation routine run.
+		var e5AllocationRoutineWindow time.Duration = 4 // Approximate duration for E5 allocation routine to run
+		now := time.Now().Local()
+		yesterday := now.Add(-24 * time.Hour) // 24 hours ago from current time
+		yesterdayAtZeroHours := time.Date(yesterday.Year(), yesterday.Month(), yesterday.Day(), 0, 0, 0, 0, time.Local)
+
+		expectedE5AllocationRoutineStartTime := yesterdayAtZeroHours.Add(time.Hour * 20)                                      // 8pm of previous day
```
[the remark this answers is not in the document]
rbull-ch: @dnair-ch has confirmed in slack that system clock is UTC - 

> "penalty-payment-api is running in EC2 instance. Default EC2 instance timezone is UTC. I don't see any terraform script changing the default timezone".

So I think we'll need to specify a location `Europe/London` to convert our configured time into.
See https://pkg.go.dev/time#LoadLocation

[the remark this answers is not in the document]
dnair-ch: As we don't have permission to login to EC2 instance, I will post a question in platform channel to confirm.

issuer_gateway/api/account_penalties.go
```diff
@@ -123,11 +123,58 @@ func convertE5Response(customerCode, companyCode string, response *e5.GetTransac
...
+			"time to live in hours":         ttlHours,
+			"cache record expiration time ": cacheRecordExpirationTime,
+			"current time":                  now,
+		})
+
+		return cacheRecordExpirationTime.Equal(time.Now()) || cacheRecordExpirationTime.Before(now)
+	} else {
+		// Penalty is marked as paid in cache, so logic to determine if cache record is stale is based on
+		// E5 allocation routine run.
+		var e5AllocationRoutineWindow time.Duration = 4 // Approximate duration for E5 allocation routine to run
```
rbull-ch, line 146: We should proabably have a Duration string in our config that we can use ParseDuration on e.g. "4h" or "3h30m"? See https://pkg.go.dev/time#Duration

issuer_gateway/api/account_penalties.go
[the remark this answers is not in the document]
dnair-ch: https://companieshouse.slack.com/archives/C05K57L4V1T/p1745837123105949?thread_ts=1745836542.566459&cid=C05K57L4V1T

issuer_gateway/api/account_penalties.go
aonubeze-ch: Yes, I intend the data to aid debugging. I will update the fields to snake_case.

issuer_gateway/api/account_penalties.go
acallaghan-ch: I would lean to making that Duration an environment variable instead of hardcoding it to 4. It will probably always be 4 but worth having the flexibility.

issuer_gateway/api/account_penalties.go
```diff
@@ -123,11 +123,58 @@ func convertE5Response(customerCode, companyCode string, response *e5.GetTransac
 }
 
 func isStale(accountPenaltiesDao *models.AccountPenaltiesDao, cfg *config.Config) bool {
-	ttlHours := cfg.AccountPenaltiesTTLHours
-	if ttlHours == 0 {
-		ttlHours = 24 // time to live defaults to 24 hours if not set in config
+	if accountPenaltiesDao.ClosedAt == nil {
```
acallaghan-ch, line 126: Would it be possible to pull out the code enclosed in the if/else into two separate methods? I've found that these large if statements have a habit of increasing in size.

issuer_gateway/api/account_penalties.go
```diff
@@ -123,11 +123,58 @@ func convertE5Response(customerCode, companyCode string, response *e5.GetTransac
...
+
+		return cacheRecordExpirationTime.Equal(time.Now()) || cacheRecordExpirationTime.Before(now)
+	} else {
+		// Penalty is marked as paid in cache, so logic to determine if cache record is stale is based on
+		// E5 allocation routine run.
+		var e5AllocationRoutineWindow time.Duration = 4 // Approximate duration for E5 allocation routine to run
+		now := time.Now().Local()
+		yesterday := now.Add(-24 * time.Hour) // 24 hours ago from current time
+		yesterdayAtZeroHours := time.Date(yesterday.Year(), yesterday.Month(), yesterday.Day(), 0, 0, 0, 0, time.Local)
+
```
acallaghan-ch, line 150: Could you just set 'yesterday at 20 hours' instead? 
yesterdayAtZeroHours := time.Date(yesterday.Year(), yesterday.Month(), yesterday.Day(), 20, 0, 0, 0, time.Local)'.

issuer_gateway/api/account_penalties.go
```diff
@@ -123,11 +123,58 @@ func convertE5Response(customerCode, companyCode string, response *e5.GetTransac
...
+		expectedE5AllocationRoutineStartTime := yesterdayAtZeroHours.Add(time.Hour * 20)                                      // 8pm of previous day
+		expectedE5AllocationRoutineEndTime := expectedE5AllocationRoutineStartTime.Add(time.Hour * e5AllocationRoutineWindow) // 12am of current day
+
+		log.Info("checking if cached account penalties have become stale", log.Data{
+			"expected E5 allocation routine start time": expectedE5AllocationRoutineStartTime,
+			"expected E5 allocation routine end time":   expectedE5AllocationRoutineEndTime,
+			"account penalties cache created at":        accountPenaltiesDao.CreatedAt,
+		})
+
+		return accountPenaltiesDao.ClosedAt.Before(expectedE5AllocationRoutineStartTime) && now.After(expectedE5AllocationRoutineEndTime)
```
acallaghan-ch, line 160: Can you add a comment here to explain what the conditions are doing? I understand now but if I was to come back to it in 6 months I'd probably struggle.

acallaghan-ch (changes requested):
Some minor suggestions

issuer_gateway/api/account_penalties.go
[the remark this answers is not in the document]
rbull-ch: I'll close this conversation following the updates from Platform and @dnair-ch - we just need to be aware the behaviour may be different later on when we move to ECS deployment and may need an additional env variable to be set to behave in the same way as mesos.

> Marc Ransome
No, the time zone will be Europe/London on all Mesos slave instances (i.e. /etc/localtime is a symbolic link to /usr/share/zoneinfo/Europe/London) as a result of the common role task which [sets the timezone](https://github.com/companieshouse/ch-service-ansible/blob/5eba89fe208fa10a88090cbe49735fc86005f44f/roles/common/tasks/main.yml#L8-L11).

> Dileepkumar Nair
We can change the time zone in ECS by setting the TZ environment variable.

issuer_gateway/api/account_penalties.go
aonubeze-ch: Yes, I will parameterise the duration, E5AllocationRoutineDuration.

issuer_gateway/api/account_penalties_test.go
```diff
@@ -52,6 +51,23 @@ var page = e5.Page{
 var e5TransactionsResponse = e5.GetTransactionsResponse{
 	Page: page,
 	Transactions: []e5.Transaction{
+		{
+			CompanyCode:          "LP",
+			LedgerCode:           "EW",
+			CustomerCode:         "12345678",
+			TransactionReference: "P1234567",
```
rbull-ch, line 58: Still got a P ref here

rbull-ch (approved):

issuer_gateway/api/account_penalties.go
acallaghan-ch: @aonubeze-ch Can you update line 140 to use: yesterday := time.Now().AddDate(0, 0, -1)
for the reason mentioned above.

rbull-ch: @aonubeze-ch  / @acallaghan-ch  I noticed there was an example of this Aaron added in a test with a `Truncate`.
I guess we might need to do that also, if you are looking more closely at this time stuff again?
See [time.Now()](https://github.com/search?q=repo%3Acompanieshouse%2Fpenalty-payment-api%20time.Now()&type=code)

acallaghan-ch (changes requested):
Update how yesterday is calculated

issuer_gateway/api/account_penalties.go
aonubeze-ch: Updated using `AddDate()`

PR #102, opened 2025-05-01
SAN-328 Upgrade Golang version to 1.24

aonubeze-ch:
* Upgrade Golang version from 1.19 to 1.24.2
* Removed dependency on [golang-base-image](https://github.com/companieshouse/golang-base-image) in the Dockerfile as version 1.24.2 of the image is not available
* Updated Dockerfile to use Golang builder 1.24.2

Depends on https://github.com/companieshouse/penalty-payment-api-core/pull/33

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,36 @@
-FROM 169942020521.dkr.ecr.eu-west-2.amazonaws.com/base/golang:1.19-bullseye-builder AS builder
+FROM golang:1.24.2-bullseye as builder
+
+ENV GOPRIVATE="github.com/companieshouse"
+
+ARG SSH_PRIVATE_KEY
+ARG SSH_PRIVATE_KEY_PASSPHRASE
+
+COPY ./bin/go_build /bin/
+
+RUN chmod +x /bin/go_build && \
+    git config --global url."git@github.com:".insteadOf https://github.com/
+
+WORKDIR /build
+
+COPY . /build/
 
 RUN /bin/go_build
 
-FROM 169942020521.dkr.ecr.eu-west-2.amazonaws.com/base/golang:debian11-runtime
+FROM debian:bullseye
+
+RUN apt-get update && \
+    apt-get install -y --allow-downgrades --no-install-recommends\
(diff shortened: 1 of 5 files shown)


acallaghan-ch (approved):

dgordon-ch (approved):

Dockerfile
```diff
@@ -1,8 +1,24 @@
...
+    git config --global url."git@github.com:".insteadOf https://github.com/
+
+WORKDIR /build
+
+COPY . /build/
 
 RUN /bin/go_build
 
-FROM 169942020521.dkr.ecr.eu-west-2.amazonaws.com/base/golang:debian11-runtime
+FROM gcr.io/distroless/base-debian11 AS runner
```
rbull-ch, line 19: @aonubeze-ch I've attempted to resolve the docker security error, but since we look to be the first team to try moving away from the companieshouse/golang-base-image then it probably needs somebody better to review what we're doing here in the Dockerfile - go upgrade

Change to `distroless/base-debian11` and `USER nonroot` to resolve `docker:S6471` Running containers as a privileged user is security-sensitive

Dockerfile
```diff
@@ -1,8 +1,24 @@
...
+ARG SSH_PRIVATE_KEY_PASSPHRASE
+
+COPY ./bin/go_build /bin/
+
+RUN chmod +x /bin/go_build && \
+    git config --global url."git@github.com:".insteadOf https://github.com/
+
+WORKDIR /build
+
+COPY . /build/
```
rbull-ch, line 15: @aonubeze-ch Any ideas about this other one?
https://sonarqube.companieshouse.gov.uk/security_hotspots?id=uk.gov.companieshouse%3Apenalty-payment-api&hotspots=e38dab3e-2e5d-4367-a252-2bec063a07a1&pullRequest=102

aonubeze-ch: I think that is fine. We have the dockerIgnore file where we white-list files we want to be copied to Docker. We only need to ensure that we white-list only files that do not contain any sensitive data.

PR #103, opened 2025-05-06
SAN-409 Remove check for penalty paid status in E5 for cache update

aonubeze-ch:
This patch removes the functionality to check for penalty paid status in E5 before updating stale cache records for account penalties. This is to simplify the cache lifecycle.

### DIFF
--- a/issuer_gateway/api/account_penalties.go
+++ b/issuer_gateway/api/account_penalties.go
@@ -42,7 +42,7 @@ func AccountPenalties(customerCode string, companyCode string, penaltyDetailsMap
 
 		if accountPenalties == nil {
 			accountPenalties = createAccountPenaltiesEntry(customerCode, companyCode, e5Response, apDaoSvc)
-		} else if paymentUpdatedInE5(e5Response, accountPenalties) {
+		} else {
 			accountPenalties = updateAccountPenaltiesEntry(customerCode, companyCode, e5Response, apDaoSvc)
 		}
 	}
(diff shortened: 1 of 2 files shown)


issuer_gateway/api/account_penalties.go
```diff
@@ -172,29 +172,6 @@ func isStale(accountPenaltiesDao *models.AccountPenaltiesDao, cfg *config.Config
...
-			return false
-		}
-	}
-
-	return true
-}
-
 func getTimeToLive(cfg *config.Config) time.Duration {
 	ttlString := cfg.AccountPenaltiesTTL
 	if ttlString == "" {
```
acallaghan-ch, line 177: I must have missed this in the last review. Why is the ttl being set to "24h"?  Should this not be just "24" as that is what the default is set to in the environment variable? Can we just check that it is an integer greater than 0 instead.

acallaghan-ch (commented):
Question around default TTL

issuer_gateway/api/account_penalties.go
aonubeze-ch: Yes, we need to add that.

dgordon-ch: Happy with either, I think we'd probably always want it on the hour of ease but having option to be more specific, down to minutes, is fine too.

dgordon-ch (approved):

issuer_gateway/api/account_penalties.go
acallaghan-ch: The env variable is called 'PPS_ACCOUNT_PENALTIES_TTL_HOURS' - so I think we won't want to be setting it to any minute increments. In the compose file it will be set to a full integer value so we don't have to cater for minutes.

acallaghan-ch: > Happy with either, I think we'd probably always want it on the hour of ease but having option to be more specific, down to minutes, is fine too.

I think that it makes it more confusing to include minutes from a setup point of view but if we are including minutes we need to update the env variable 'PPS_ACCOUNT_PENALTIES_TTL_HOURS' as it's not just hours any longer.

aonubeze-ch: I am happy to update the env variable. I would need to update it also in chs-configs and docker-chs-development.

acallaghan-ch (approved):

PR #151, opened 2025-07-03
SAN-383 Improve logging

aonubeze-ch:
Added some more verbose logs to aid debugging.


common/dao/mongo.go
```diff
@@ -251,7 +253,7 @@ func (m *MongoAccountPenaltiesService) UpdateAccountPenaltyAsPaid(customerCode s
 
 // UpdateAccountPenalties updates the created_at, closed_at and data fields of an existing document
 func (m *MongoAccountPenaltiesService) UpdateAccountPenalties(dao *models.AccountPenaltiesDao) error {
-	log.Info("updating existing document in account_penalties collection", log.Data{
+	log.Debug("updating existing document in account_penalties collection", log.Data{
```
acallaghan-ch, line 256: Same for this.

common/dao/mongo.go
```diff
@@ -309,8 +311,8 @@ func (m *MongoPayableResourceService) SaveE5Error(customerCode, payableRef strin
 	update := bson.D{
 		{
 			Key: "$set", Value: bson.D{
```
rbull-ch, line 313: Did we need to reformat these lines?  It looks odd to me with the brace alignment.

Wondering if this would be any better?
```
			Key: "$set", Value: bson.D{{Key: "e5_command_error", Value: string(action)}},
```

issuer_gateway/api/account_update.go
```diff
@@ -39,17 +41,28 @@ func UpdateIssuerAccountWithPenaltyPaid(payableResourceService *services.Payable
...
 	}
 
 	// three http requests are needed to mark a transactions as paid. The process is 1) create the payment, 2) authorise
 	// the payments and finally 3) confirm the payment. if anyone of these fails, the company account will be locked in
 	// E5. Finance have confirmed that it is better to keep these locked as a cleanup process will happen naturally in
 	// the working day.
+	logData := log.Data{
+		"company_code":  companyCode,
+		"customer_code": resource.CustomerCode,
+		"penaltyRef":    transactions[0].TransactionReference,
```
rbull-ch, line 59: This data param should probably be in snake_case also, so `penalty_ref`?

common/dao/mongo.go
rbull-ch: I looked before I commented, my difftool shows the same

image

rbull-ch (approved):

PR #161, opened 2025-07-24
SAN-428: Add retry for transient E5 errors

acallaghan-ch:
Add retry for transient E5 errors
Using the avast Go library to manage the retries.
Also included method for the retry topic that will not be called until the retry consumer is implemented

Related Config: https://github.com/companieshouse/docker-chs-development/pull/1867

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,30 +13,33 @@ In order to run this API locally you will need to install the following:
 2. Build the executable: `make build`
 
 ## Configuration
-| Variable                                   | Default | Description                                                                 | Config Location                                                           |
-|:-------------------------------------------|:-------:|:----------------------------------------------------------------------------|---------------------------------------------------------------------------|
-| `API_KEY`                                  |   `-`   | API Key to call payments API                                                | Terraform Vault - To update, please create platform request               |
-| `E5_USERNAME`                              |   `-`   | E5 API Username                                                             | Terraform Vault - To update, please create platform request               |
-| `MONGODB_URL`                              |   `-`   | The mongo db connection string                                              | Terraform Vault - To update, please create platform request               |
-| `E5_API_URL`                               |   `-`   | E5 API Address                                                              | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `PPS_MONGODB_DATABASE`                     |   `-`   | The database name to connect to e.g. `financial_penalties`                  | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `PPS_MONGODB_PAYABLE_RESOURCES_COLLECTION` |   `-`   | The collection name e.g. `payable_resources`                                | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `PPS_MONGODB_ACCOUNT_PENALTIES_COLLECTION` |   `-`   | The collection name e.g. `account_penalties`                                | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `PPS_ACCOUNT_PENALTIES_TTL`                |   `-`   | Account penalties cache time to live  e.g. `24h`                            | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `KAFKA_BROKER_ADDR`                        |   `_`   | Kafka Broker Address                                                        | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `KAFKA_ZOOKEEPER_ADDR`                     |   `_`   | Kafka Zookeeper Address                                                     | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `SCHEMA_REGISTRY_URL`                      |   `_`   | Schema Registry URL                                                         | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `EMAIL_SEND_TOPIC`                         |   `_`   | Kafka topic to send emails e.g. email-send                                  | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `PENALTY_PAYMENTS_PROCESSING_TOPIC`        |   `_`   | Kafka topic to process penalty payments to e.g. penalty-payments-processing | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `FEATURE_FLAG_PAYMENTS_PROCESSING_ENABLED` |   `_`   | If the payments processing Kafka implementation is enabled                  | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `API_URL`                                  |   `_`   | The application endpoint for the API, for go-sdk-manager integration        | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `PAYMENTS_API_URL`                         |   `_`   | The base path for the payments API, for go-sdk-manager integration          | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `CHS_URL`                                  |   `_`   | CHS URL                                                                     | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `WEEKLY_MAINTENANCE_START_TIME`            |   `_`   | Start time of weekly maintenance e.g. `0700`                                | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
-| `WEEKLY_MAINTENANCE_END_TIME`              |   `_`   | End time of weekly maintenance e.g. `0730`                                  | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -13,30 +13,33 @@ In order to run this API locally you will need to install the following:
...
+| `EMAIL_SEND_TOPIC`                            |   `_`   | Kafka topic to send emails e.g. email-send                                  | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
+| `PENALTY_PAYMENTS_PROCESSING_TOPIC`           |   `_`   | Kafka topic to process penalty payments to e.g. penalty-payments-processing | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
+| `PENALTY_PAYMENTS_PROCESSING_MAX_RETRIES`     |   `_`   | The max retry attempts for transient errors to e.g. 3                       | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
+| `PENALTY_PAYMENTS_PROCESSING_RETRY_DELAY`     |   `_`   | The delay in seconds between retry attempts for transient errors to e.g. 1  | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
+| `PENALTY_PAYMENTS_PROCESSING_RETRY_MAX_DELAY` |   `_`   | The maximum delay time between retires for transient errors                 | ecs-service-configs-dev(CIDEV) / ecs-service-configs-prod (STAGING/LIVE)  |
```
rbull-ch, line 33: Typo - `retires` should be `retries`.
Also does the description need to say delay in seconds like the `PENALTY_PAYMENTS_PROCESSING_RETRY_DELAY`?

handlers/finance_payment.go
```diff
@@ -33,26 +42,162 @@ func (p PenaltyFinancePayment) ProcessFinancialPenaltyPayment(penaltyPayment mod
 	}
 	log.Info("Financial penalty payment processing started", logContext)
 
-	createPaymentError, createPaymentSuccess := createPayment(penaltyPayment, p.PayableResourceDaoService, p.E5Client, e5PaymentID)
-	if !createPaymentSuccess {
-		return createPaymentError
+	payProcessingEnabled := getPaymentsProcessingEnabled()
```
rbull-ch, line 45: I don't think this `payProcessingEnabled` is needed in the 5 places with all the if/else statements as we check the config FF is enabled when the application starts the consumer in [main.go#L56](https://github.com/companieshouse/penalty-payment-api/blob/0b2dd08c4fba0aa4d9775b732234d4b34dd630a6/main.go#L56)

handlers/finance_payment.go
```diff
@@ -33,26 +42,162 @@ func (p PenaltyFinancePayment) ProcessFinancialPenaltyPayment(penaltyPayment mod
...
 }
 
-func createPayment(penaltyPayment models.PenaltyPaymentsProcessing, payableResourceDaoService dao.PayableResourceDaoService, client e5.ClientInterface, e5PaymentID string) (createPaymentError error, createPaymentSuccess bool) {
+func isAfterNextDayMidnight(createdAt string) bool {
+	parsed, _ := time.Parse(time.RFC3339, createdAt)
+	nextDayMidnight := time.Date(parsed.Year(), parsed.Month(), parsed.Day()+1, 0, 0, 0, 0, parsed.Location())
+	return time.Now().After(nextDayMidnight)
+}
+
+func withRetry(operationName string, fn func() error) error {
```
rbull-ch, line 150: The first parameter could should probably just use the type `e5.Action`.
No need for the string thing here.

handlers/finance_payment.go
```diff
@@ -33,26 +42,162 @@ func (p PenaltyFinancePayment) ProcessFinancialPenaltyPayment(penaltyPayment mod
...
+	var attemptsStr = cfg.PenaltyPaymentsProcessingMaxRetries
+	var attempts = uint(3)
+	if s, err := strconv.ParseUint(attemptsStr, 10, 32); err == nil {
+		attempts = uint(s)
+	}
+	return attempts
+}
+
+func getDelay(cfg *config.Config) time.Duration {
+	var delayStr = cfg.PenaltyPaymentsProcessingRetryDelay
```
rbull-ch, line 182: Can we have the duration specified in the config, which would make it more flexible, simpler to understand
e.g. `1s` and the code would just use `time.ParseDuration`?

acallaghan-ch: I thought about that but the reason I didn't do it was to force the duration to be in seconds.
I don't see us ever wanted it anything but in seconds, and having it configurable to minutes or hours seems wrong to me.

handlers/finance_payment.go
```diff
@@ -33,26 +42,162 @@ func (p PenaltyFinancePayment) ProcessFinancialPenaltyPayment(penaltyPayment mod
...
+		saveE5Error(penaltyPayment, p.PayableResourceDaoService, err, e5PaymentID, e5.ConfirmAction)
+		return err
+	}
+
+	log.Info("Financial penalty payment processing successful", logContext)
+	return nil
+}
+
+// This method should be called by the retry topic - currently not being used
+func (p PenaltyFinancePayment) ProcessFinancialPenaltyPaymentRetryTopic(penaltyPayment models.PenaltyPaymentsProcessing, e5PaymentID string) error {
```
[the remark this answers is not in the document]
acallaghan-ch: I don't think we want the 3 retries thing to happen on the non-transient errors. The calls would only happen once. 
Also, on the retry we want a way to mark the message as processed on the topic to stop any further retry attempts which is logic that is not needed in the transient retries. It keeps it much cleaner for me.

handlers/finance_payment.go
```diff
@@ -33,26 +42,162 @@ func (p PenaltyFinancePayment) ProcessFinancialPenaltyPayment(penaltyPayment mod
...
 	}
 
-	authorisePaymentError, authorisePaymentSuccess := authorisePayment(penaltyPayment, p.PayableResourceDaoService, p.E5Client, e5PaymentID)
-	if !authorisePaymentSuccess {
-		return authorisePaymentError
+	err = authorisePayment(penaltyPayment, p.E5Client, e5PaymentID)
+	if err != nil {
+		saveE5Error(penaltyPayment, p.PayableResourceDaoService, err, e5PaymentID, e5.AuthoriseAction)
+		// this should now be marked as processed on the retry topic as the create succeeded - we are not tracking state
+		return err
```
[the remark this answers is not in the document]
acallaghan-ch: I have just missed the err return on the authorise - I will put in, but I agree that actually they should return nil.
Regarding state, it means that we are not tracking the state in terms of:
- create has succeeded but authorise has failed, so lets try authorise again 
- create has succeeded, authorise has succeeded, but confirm had failed so lets try the confirm again
as discussed in the workshop.

handlers/create_payable_resource.go
```diff
@@ -18,7 +18,6 @@ import (
 	"github.com/companieshouse/penalty-payment-api/penalty_payments/transformers"
 )
 
-var getCompanyCodeFromTransaction = utils.GetCompanyCodeFromTransaction
```
dgordon-ch, line 21: Is this just cleanup, no longer used?

handlers/finance_payment.go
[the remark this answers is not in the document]
dgordon-ch: If it's catering to different functionality then agreed at keeping separate.

handlers/finance_payment.go
acallaghan-ch: I forgot that we had separate logic for the non kafka stuff, I'll remove.

handlers/finance_payment.go
```diff
@@ -33,26 +42,163 @@ func (p PenaltyFinancePayment) ProcessFinancialPenaltyPayment(penaltyPayment mod
...
+	delay := getDelay(cfg)
+	maxDelay := getMaxDelay(cfg)
+
+	return retry.Do(
+		fn,
+		retry.Attempts(attempts),
+		retry.Delay(delay),
+		retry.MaxDelay(maxDelay),
+		retry.OnRetry(func(n uint, err error) {
+			log.Info("Penalty payment processing retry attempt failed: " + string(action))
```
rbull-ch, line 168: Even here - I don't think you need the string bit for the log when I looked at this yesterday

handlers/finance_payment.go
```diff
@@ -33,26 +42,163 @@ func (p PenaltyFinancePayment) ProcessFinancialPenaltyPayment(penaltyPayment mod
...
+func isAfterNextDayMidnight(createdAt string) bool {
+	parsed, _ := time.Parse(time.RFC3339, createdAt)
+	nextDayMidnight := time.Date(parsed.Year(), parsed.Month(), parsed.Day()+1, 0, 0, 0, 0, parsed.Location())
+	return time.Now().After(nextDayMidnight)
+}
+
+func withRetry(action e5.Action, fn func() error) error {
+	cfg, err := getConfig()
+	if err != nil {
+		err = fmt.Errorf("error getting config for kafka message production: [%v]", err)
```
rbull-ch, line 138: I don't think this error makes sense saying `production` unless it's doing something with the topic retry message stuff?

handlers/finance_payment.go
```diff
@@ -33,26 +42,163 @@ func (p PenaltyFinancePayment) ProcessFinancialPenaltyPayment(penaltyPayment mod
...
+		// this should now be marked as processed on the retry topic
+		return nil
+	}
+
+	var err error
+
+	err = createPayment(penaltyPayment, p.E5Client, e5PaymentID)
+	if err != nil {
+		// Should go back on the topic with the number of attempts updated
+		return nil
```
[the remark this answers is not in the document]
acallaghan-ch: Yeah, I think that is something to be tackled in the next ticket that is dealing with the retry topic.

handlers/finance_payment.go
acallaghan-ch: This is now gone with the new paramter being used.

handlers/finance_payment.go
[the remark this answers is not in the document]
acallaghan-ch: My understanding is that if we get any error back from E5 in the initial call, then we consider it a transient error.

PR #231, opened 2025-11-24
SAN-783: Remove existing config files and use new config coming from …

acallaghan-ch:
Remove the previous configuration files and use the new configuration files coming from api-core
Update all associated unit tests


issuer_gateway/private/payable_status_provider.go
```diff
@@ -21,18 +22,19 @@ const (
...
+		e5Transactions []models.AccountPenaltiesDataDao, penaltyTypes map[string]map[string]finance_config.FinancePenaltyTypeConfig,
+		cfg *config.Config) string
 }
 
 type DefaultPayableStatusProvider struct{}
 
 func (provider *DefaultPayableStatusProvider) GetPayableStatus(transactionType string, e5Transaction *models.AccountPenaltiesDataDao, closedAt *time.Time,
-	e5Transactions []models.AccountPenaltiesDataDao, allowedTransactionsMap *models.AllowedTransactionMap, cfg *config.Config) string {
+	e5Transactions []models.AccountPenaltiesDataDao, penaltyTypes map[string]map[string]finance_config.FinancePenaltyTypeConfig, cfg *config.Config) string {
 	if types.Penalty.String() == transactionType {
```
rbull-ch, line 33: For the new finance_payable_penalties config we should strictly be checking under the finance_payment.transaction_type and finance_payment.transaction_sub_types list for a payable penalty.
e.g.
* finance_payable_penalties.LATE_FILING.finance_payment.transaction_type
* finance_payable_penalties.LATE_FILING.finance_payment.transaction_sub_types

Also since we now have a way to disable a transaction_subtype in the new finance config, we should check that it is not disabled e.g.
* finance_penalty_types.1.A1.disabled

penalty_payments/service/payment_details.go
```diff
@@ -18,8 +18,8 @@ type PaymentDetailsService struct {
 
 // GetPaymentDetailsFromPayableResource transforms a PayableResource into its corresponding Payment details resource
 func (service *PaymentDetailsService) GetPaymentDetailsFromPayableResource(req *http.Request,
-	payable *models.PayableResource, penaltyDetails config.PenaltyDetails) (*models.PaymentDetails, error) {
-	paymentDetails := transformers.PayableResourceToPaymentDetails(payable, penaltyDetails)
+	payable *models.PayableResource, ppConfig finance_config.FinancePayablePenaltyConfig) (*models.PaymentDetails, error) {
+	paymentDetails := transformers.PayableResourceToPaymentDetails(payable, ppConfig)
```
rbull-ch: Can we just use the `PaymentCostConfig` here as a param instead?

handlers/payment_details.go
```diff
@@ -37,14 +37,14 @@ func HandleGetPaymentDetails(w http.ResponseWriter, req *http.Request) {
...
-	log.DebugC(requestId, "penalty details", log.Data{"penaltyDetails": penaltyDetails})
+	payablePenaltyConfig := penaltyConfig.PayablePenalties[penaltyRefType]
+	log.DebugC(requestId, "penalty details", log.Data{"payablePenaltyConfig": payablePenaltyConfig})
 
 	// Get the payment details from the payable resource
 	logContext := log.Data{"customer_code": payableResource.CustomerCode, "payable_ref": payableResource.PayableRef}
 	log.InfoC(requestId, "getting payment details", logContext)
 	paymentDetails, err := paymentDetailsService.GetPaymentDetailsFromPayableResource(req,
-		payableResource, penaltyDetails)
+		payableResource, payablePenaltyConfig)
```
rbull-ch: Can we just use the `PaymentCostConfig` here as a param instead?

penalty_payments/service/email_service.go
```diff
@@ -169,9 +169,9 @@ func prepareEmailKafkaMessage(emailSendSchema avro.Schema, payableResource model
 	messageID := "<" + payableResource.PayableRef + "." + strconv.Itoa(util.Random(0, 100000)) + "@companieshouse.gov.uk>"
 
 	emailSendMessage := models.EmailSend{
-		AppID:        penaltyConfig.PenaltyDetailsMap.Details[penaltyRefType].EmailReceivedAppId,
+		AppID:        penaltyConfig.PayablePenalties[penaltyRefType].EmailSend.AppId,
```
rbull-ch: Please can you declare a variable for the `EmailSendConfig` that can be re-used in the 2 places here?

handlers/payment_details.go
```diff
@@ -37,14 +37,14 @@ func HandleGetPaymentDetails(w http.ResponseWriter, req *http.Request) {
 
 	penaltyConfig := configctx.FromContext(req.Context())
 
-	penaltyDetails := penaltyConfig.PenaltyDetailsMap.Details[penaltyRefType]
-	log.DebugC(requestId, "penalty details", log.Data{"penaltyDetails": penaltyDetails})
+	payablePenaltyConfig := penaltyConfig.PayablePenalties[penaltyRefType]
+	log.DebugC(requestId, "penalty details", log.Data{"payablePenaltyConfig": payablePenaltyConfig})
```
rbull-ch: I think this debug log could be improved if you change the param below to use the payment_cost object.  That's all we're interested in here, so we could change the message to say "payment cost" and the data to have `payment_cost` like seen in the yaml.
We normally use snake_case.

issuer_gateway/private/payable_status_provider.go
acallaghan-ch: Good point but I would prefer to stick new functionality in a new ticket and a more robust test suite can be created.
I'll pick the ticket up - https://companieshouse.atlassian.net/browse/SAN-785
@rbull-ch Sound ok?

middleware/penalty_config.go
rbull-ch, line 1: Since the old yaml config is being removed in this PR, then I think you'll also need to remember to visit the files with lines matching `assets` like in commit 74d42f8bced9431ffa22b4156fcc12f707bdb900 e.g. .dockerignore / Makefile

acallaghan-ch: I removed the entry from _.dockerignore_ and I had a check there and there is no reference to 'assets' in the Makefile an longer as we are using the FileSystem.

rbull-ch: Thanks, yep I was confused why I couldn't see it in the changelist.  I remembered us doing something about that recently. Looks ok, my mistake!

version
```diff
@@ -1 +1 @@
-1.22
+1.23
```
rbull-ch: Unfortunately, this might look daft now when it comes to merging the shared dev branch [feature/SAN-737-Penalty-Subtype-improvements](https://github.com/companieshouse/penalty-payment-api/tree/feature/SAN-737-Penalty-Subtype-improvements) into main at some point as we will be jumping versions without any tagged release.
So maybe it should remain as 1.22? Since last tagged release is https://github.com/companieshouse/penalty-payment-api/releases/tag/1.21.1

rbull-ch (approved):
Approved with minor comment around versioning - https://github.com/companieshouse/penalty-payment-api/pull/231/files#r2556850564

version
acallaghan-ch: Dead on, I'll revert the version

aonubeze-ch (approved):

